PR #43, opened 2020-04-23
Backup improvements for restoring without a master

feikesteenbergen:
By default, Patroni will not restore a backup if no master can be found.
However, it may be a totally valid usecase to pause your deployment.

Therefore, we will allow bootstrapping to be done using the backup even
if no master is currently running.

Second, the race between the pgBackRest wrapper and initdb seemed to
have been a bit to tight, so allow 30 seconds instead of 10 seconds to
pass before bailing out on triggering a backup.

Third, a misconfiguration, as --stanza=demo never existed. As the
command was only used in the thus far never supported mode of bootstrap
without master, it never was an issue.

### DIFF
--- a/scripts/augment_patroni_configuration.py
+++ b/scripts/augment_patroni_configuration.py
@@ -22,13 +22,14 @@ postgresql:
     command: '/usr/bin/pgbackrest --stanza=poddb --delta restore --log-level-stderr=info'
     keep_data: True
     no_params: True
+    no_master: 1
 bootstrap:
   dcs:
     postgresql:
       recovery_conf:
         recovery_target_timeline: latest
         standby_mode: 'on'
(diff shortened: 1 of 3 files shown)


scripts/augment_patroni_configuration.py
```diff
@@ -22,13 +22,14 @@
     command: '/usr/bin/pgbackrest --stanza=poddb --delta restore --log-level-stderr=info'
     keep_data: True
     no_params: True
+    no_master: 1
```
iand, line 25: I know 1 is a synonym for True but the previous two flags use True so perhaps we should here

feikesteenbergen: I agree with the ugliness, I'm following the documentation here:

https://patroni.readthedocs.io/en/latest/replica_bootstrap.html#building-replicas

I might throw in a PR upstream

PR #45, opened 2020-05-01
Feike/include timescale prometheus

feikesteenbergen:


Dockerfile
```diff
@@ -152,6 +152,22 @@ RUN if [ ! -z "${PG_PROMETHEUS}" ]; then \
         done; \
     fi
 
+ARG TIMESCALE_PROMETHEUS=
+# build and install the pg_prometheus extension
+RUN if [ ! -z "${TIMESCALE_PROMETHEUS}" ]; then \
+        mkdir -p /build \
+        && git clone https://github.com/timescale/timescale-prometheus /build/timescale_prometheus \
+        && set -e \
+        && for pg in ${PG_VERSIONS}; do \
```
feikesteenbergen, line 161: This Docker image contains multiple PostgreSQL binaries, 11 and 12 currently.
The idiomatic way we currently install is by looping over the versions.
I could rip it out, but then with 13 support (I'm hoping we'll have that soon) we'll have to put it back in.

Makefile
```diff
@@ -30,6 +30,7 @@ TIMESCALEDB_BUILDER_URL?=$(TIMESCALEDB_IMAGE):builder-$(PGVERSION)
 TIMESCALEDB_RELEASE_URL?=$(TIMESCALEDB_IMAGE):$(TAG)-$(PGVERSION)
 TIMESCALEDB_LATEST_URL?=$(TIMESCALEDB_IMAGE):latest-$(PGVERSION)
 PG_PROMETHEUS?=0.2.2
+TIMESCALE_PROMETHEUS?=master
```
feikesteenbergen, line 33: Would be great if we could point to a tagged release here instead of master, even if that tag is a development version

Dockerfile
blagojts: Will having both/three binaries make it easier to run upgrade scripts between versions? Like upgrade image version, but still start PG11, ???, shutdown, start PG12, run upgrade. Or the idea is to utilize some other mechanism?

Makefile
```diff
@@ -30,6 +30,7 @@ TIMESCALEDB_BUILDER_URL?=$(TIMESCALEDB_IMAGE):builder-$(PGVERSION)
 TIMESCALEDB_RELEASE_URL?=$(TIMESCALEDB_IMAGE):$(TAG)-$(PGVERSION)
 TIMESCALEDB_LATEST_URL?=$(TIMESCALEDB_IMAGE):latest-$(PGVERSION)
 PG_PROMETHEUS?=0.2.2
```
blagojts, line 32: We should deprecate this

blagojts: or completely remove

Dockerfile
feikesteenbergen: To upgrade, it is required to have binaries of the current `$PGDATA` directory and the newer `$PGDATA` directory, so yes, that's the main reason to stick multiple binaries in.

Makefile
feikesteenbergen: Deprecated it by setting it by default to *not* build.

sgichohi:
Hey @feikesteenbergen did we build this, I can roll it out if we did.

PR #47, opened 2020-06-05
Include debug symbols for PostgreSQL

feikesteenbergen:


Dockerfile
```diff
@@ -125,7 +125,7 @@ RUN TS_VERSIONS=$(curl "https://api.github.com/repos/${GITHUB_REPO}/releases" \
             if [ ${pg} -ge 12 ] && dpkg --compare-versions ${ts} lt 1.7.0; then echo "Skipping: TimescaleDB ${ts} is not supported on PostgreSQL ${pg}" && continue; fi \
             && cd /build/timescaledb && git reset HEAD --hard && git checkout ${ts} \
             && rm -rf build \
-            && PATH="/usr/lib/postgresql/${pg}/bin:${PATH}" ./bootstrap -DREGRESS_CHECKS=OFF -DPROJECT_INSTALL_METHOD="${INSTALL_METHOD}"${OSS_ONLY} \
+            && PATH="/usr/lib/postgresql/${pg}/bin:${PATH}" ./bootstrap -DCMAKE_BUILD_TYPE=RelWithDebInfo -DREGRESS_CHECKS=OFF -DPROJECT_INSTALL_METHOD="${INSTALL_METHOD}"${OSS_ONLY} \
```
niksajakovljevic, line 128: I believe that this will add debug symbols into binary. I'd rather go with having debug symbols stored in a separate file. This is mostly due to the fact that having debug symbols might affect performances and maybe some other weirdness...
If we use `gcc` then I guess generating external debug symbols require some additional steps...

feikesteenbergen: @niksajakovljevic I'm by no means an expert here, however this answer seems to suggest that the only overhead is storage footprint, not performance:

https://stackoverflow.com/a/8676610/2459656

gayyappan: Yes, I favor stripping the symbols as well.

feikesteenbergen: I'm going to merge this one, as we need the debug symbols to get useful core dumps.
If later one we want to optimize this, let's do that.

PR #51, opened 2020-06-16
Include custom after install/update scripts

feikesteenbergen:
These custom scripts - if configured - can be used to run after a
pgextwlist allowed extension is created.

https://github.com/dimitri/pgextwlist#custom-scripts


scripts/timescale/after-update.sql
```diff
@@ -0,0 +1 @@
+after-create.sql
```
feikesteenbergen, line 1: It's a symlink ;-)

PR #55, opened 2020-06-29
Add grants to internal timescale schemas

sgichohi:
To enable seamless restores for our users on Forge, we need to grant access to internal timescale schema and tables to the database owner.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .build*
+.idea/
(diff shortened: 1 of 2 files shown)


sgichohi:
Tested this on a local instance I had and it worked well. I wish we had instructions on how to test the images on here as well, I got stumped, but learned a lot.

PR #67, opened 2020-09-29
Feike/promscale

feikesteenbergen:
# Include promscale_extension in the Docker Image
    
The timescale_prometheus_extra is renamed to the promscale_extension.
For now, we include both versions.

# Reduce size of Docker Image

These directories are only required during building the Rust
componenents.


sgichohi (approved):

Dockerfile
```diff
@@ -170,6 +171,23 @@ RUN if [ ! -z "${TIMESCALE_PROMETHEUS}" ]; then \
         done; \
     fi
 
+ARG TIMESCALE_PROMSCALE_EXTENSION=
+# build and install the promscale_extension extension
+RUN if [ ! -z "${TIMESCALE_PROMETHEUS}" ]; then \
```
cevian, line 176:
```suggestion
RUN if [ ! -z "${TIMESCALE_PROMSCALE_EXTENSION}" ]; then \
```

Makefile
```diff
@@ -34,6 +34,7 @@ TIMESCALEDB_RELEASE_URL?=$(TIMESCALEDB_IMAGE):$(TAG)
 TIMESCALEDB_LATEST_URL?=$(TIMESCALEDB_IMAGE):latest
 PG_PROMETHEUS?=
 TIMESCALE_PROMETHEUS?=0.1.0-beta.4
+TIMESCALE_PROMSCALE_EXTENSION?=cd691614
```
cevian, line 37:
```suggestion
TIMESCALE_PROMSCALE_EXTENSION?=0.1.1
```

cevian (approved):

Dockerfile
```diff
@@ -170,6 +171,23 @@ RUN if [ ! -z "${TIMESCALE_PROMETHEUS}" ]; then \
         done; \
     fi
 
+ARG TIMESCALE_PROMSCALE_EXTENSION=
+# build and install the promscale_extension extension
+RUN if [ ! -z "${TIMESCALE_PROMSCALE_EXTENSION}" ]; then \
+        curl https://sh.rustup.rs -sSf | bash -s -- -y \
```
sgichohi, line 177: wonder if we could save time by not having to redo this now while the other extension lives

feikesteenbergen: Takes only a few seconds, it does not redownload all required tools.

PR #75, opened 2020-12-21
Include TimescaleDB 2.0.0

feikesteenbergen:
For completeness, we also include 2.0.0-rc3 and 2.0.0-rc4, as we would
otherwise force those users that upgraded to a Release Candidate to
immediately upgrade after this Docker Image is used.


feikesteenbergen:
This is expected to fail until `2.0.0` is actually tagged in the upstream repo though.

niksajakovljevic (approved):

PR #92, opened 2021-03-02
Use GitHub actions for CI/CD

feikesteenbergen:
We (Timescale) have made the decision to move to GitHub actions as much
as possible. This repository - even though public - is no exception to
that and therefore now also gets the GitHub Actions treatment.

The move itself is not that hard, however dependencies are involved
which is why this commit is actually quite big. As we no longer use
GitLab, we also want to remove the dependency of our Docker Images being
stored in a GitLab Docker registry.

That is why this commit moves the Docker Images to a (private) ECR
registry, as well as the publicly available timescale Docker Hub
repository.

Up to this point, we published our images on Docker Hub on the
timescaledev/timescaledb-ha repository. As many have pointed out the
`dev` part of that repository may cause confusion as these images images
that can be used in production and should not be considered `dev`
material.
Therefore, the images now get pushed to the timescale/timescaledb-ha
repository.
To aid in the repository name switch, the immutable images on
timescaledev/timescaledb-ha have been republished on timescale/timescaledb-ha.

This having become such a big change anyway, it felt the time was ripe
to do some cleanup of the logic of the Makefile - especially as the
Makefile and the CI/CD method are tightly coupled.
Whereas before this commit, some Make targets would loop, we now fully
delegate the loops (GH Actions: matrix) to CI/CD. In other words,
`make` will always build 1 set of images (say PostgreSQL 12),
but no longer builds multiple sets of images.
This also allows CI/CD to parallelize the building of different images,
which (in theory) should speed up the build process.

To aid in helping other tools determine what is in the Docker Image, the
newly built Docker Images will also be labelled with all the available
timescaledb versions that are included, not just the default version.


feikesteenbergen:
Seems to work:

https://github.com/timescale/timescaledb-docker-ha/runs/2039568983?check_suite_focus=true#step:8:22

Makefile
```diff
@@ -7,195 +7,166 @@ PG_VERSIONS?=12 11
...
-	export PGVERSION=$$(docker inspect $(CICDIMAGE) | jq '.[0]."ContainerConfig"."Labels"."com.timescaledb.image.postgresql.version"' -r) \
-	&& export TSPATCH=$$(docker inspect $(CICDIMAGE) | jq '.[0]."ContainerConfig"."Labels"."com.timescaledb.image.timescaledb.version"' -r) \
-	&& export TSMINOR=$${TSPATCH%.*} \
-	&& for variant in pg$(PG_MAJOR)$(POSTFIX)-latest pg$(PG_MAJOR)-ts$${TSMINOR}$(POSTFIX)-latest pg$${PGVERSION}-ts$${TSMINOR}$(POSTFIX)-latest pg$${PGVERSION}-ts$${TSPATCH}$(POSTFIX)-latest; \
-		do \
-			docker tag $(CICDIMAGE) $(PUBLISH_REPOSITORY):$${variant} \
-			&& docker push $(PUBLISH_REPOSITORY):$${variant} || exit 1; \
+# Our method of finding a patch version is quite brute force (`docker pull image`), 
+# however, we should not be publishing that often, so we take the hit for now.
+publish: publish-mutable publish-next-patch-version
```
feikesteenbergen, line 123: Out of date comment, we no longer pull, we do http calls checking for `200` or `404`

Makefile
```diff
@@ -7,195 +7,166 @@ PG_VERSIONS?=12 11
...
+# These variables have to do with what software we pull in from github for timescaledb
+GITHUB_USER?=
+GITHUB_TOKEN?=
+GITHUB_REPO?=timescale/timescaledb
+GITHUB_TAG?=master
+
+# We need dynamic variables here, that is why we do not use $(shell awk ...)
+VAR_PGMINOR="$$(awk -F '=' '/postgresql.version=/ {print $$2}' $(VAR_VERSION_INFO))"
+VAR_TSMINOR="$$(awk -F '=' '/timescaledb.version=/ {print $$2}' $(VAR_VERSION_INFO))"
+VAR_TSMAJOR="$$(awk -F '[.=]' '/timescaledb.version=/ {print $$2 "." $$3}' $(VAR_VERSION_INFO))"
```
feikesteenbergen, line 42:
```suggestion
VAR_TSMAJOR="$$(awk -F '[.=]' '/timescaledb.version=/ {print $$2}' $(VAR_VERSION_INFO))"
```

Makefile
```diff
@@ -7,195 +7,164 @@ PG_VERSIONS?=12 11
...
+			done; \
+			exit 0; \
+		elif [ "$${DOCKER_HUB_HTTP_CODE}" = "200" ]; then \
+			echo "$${IMMUTABLE_TAG} already exists, incrementing patch number"; \
+		else \
+			echo "Unexpected HTTP return code: $${DOCKER_HUB_HTTP_CODE}"; \
+			exit 1 ;\
+		fi \
+	done
+
```
alexeyklyukin, line 152: very unlikely to happen, but should we complain if we reached past 100 without coming across non-existing image?

feikesteenbergen: Yeah, let's add it. > 100 would be a bug probably, but better to complain than to succeed silently.

alexeyklyukin (approved):
Other than some minor nits, it looks good to me!

README.md
```diff
@@ -62,29 +60,15 @@ The release commit should be tagged with a signed tag:
...
-They will be written under quite a few aliases, for example, for PostgreSQL 12.3 and Timescale 1.7.1, the following images will be built and pushed/overwritten:
-
-- timescaledev/timescaledb-ha:pg12-latest
-- timescaledev/timescaledb-ha:pg12-ts1.7-latest
-- timescaledev/timescaledb-ha:pg12.4-ts1.7-latest
-- timescaledev/timescaledb-ha:pg12.4-ts1.7.2-latest
+- timescale/timescaledb-ha:pg12-latest
+- timescale/timescaledb-ha:pg12-ts2.0-latest
+- timescale/timescaledb-ha:pg12.6-ts2.0-latest
+- timescale/timescaledb-ha:pg12.6-ts2.0.1-latest
```
alexeyklyukin, line 71:
```suggestion
- timescale/timescaledb-ha:pg12.6-ts2.0.1-latest

**Note** the `-latest` suffix here indicates the latest docker build, not the latest commit. In particular, images built from a tag will be published with the `-latest` suffix in addition to the tag-based suffix.
```

sgichohi (approved):
Not a blocker, but really simple instructions for forge related tasks could be cool. And maybe an approval from the DB team or Analytics team would check for usability on their end.

.github/workflows/publish_image.yaml
```diff
@@ -0,0 +1,77 @@
+name: Publish image
+
+on:
+  push:
+    tags:
+    - 'test*'
+    - 'v*'
```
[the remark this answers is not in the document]
feikesteenbergen: Ah, well spotted, need to remove that one.

feikesteenbergen:
> Not a blocker, but really simple instructions for forge related tasks could be cool

I'll merge this one first, and create follow up PR's to not muddy the waters too much

PR #95, opened 2021-03-09
Allow extensions to be added to a live container

feikesteenbergen:
By default, the directories supporting extensions and their files are
owned by root, and are readable by PostgreSQL.

This is (in general) a good thing: It ensures that these files cannot be
updated, not even if an end-user has access to the `superuser` of the
PostgreSQL cluster.

This does however prevent *any* live patching of these directories being
done on a live container, the container itself is running as a
non-privileged container in most environments, and therefore no files
can be added. This means that any change - even adding an extension
version - requires the container to be stopped and started using a
different Docker Image.

By allowing the *addition only* of new files to the directories by the
PostgreSQL user, we open up some Development opportunities, without
causing to many surprises for running processes.

This commit does the following:

1. make `postgres` own specific extension related directories
2. the existing files in the directory are still fully owned by the root user
3. the sticky bit is set on the directory

This ensures the following:

- *new* files can be created, modified and deleted by the `postgres` user
- *existing* files (in the image) cannot be deleted or modified by postgres

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -224,6 +224,21 @@ RUN chmod +x /usr/local/bin/docker-entrypoint.sh
 RUN ln -s /usr/bin/timescaledb-tune /usr/local/bin/timescaledb-tune
 RUN ln -s /usr/local/bin/docker-entrypoint.sh /docker-entrypoint.sh
 
+# Allow Adding Extensions allows *new* files to be created, so that extensions can be added to a running container.
+# Existing files are still owned by root and have their sticky bit (the 1 in the 1775 permission mode) set,
+# and therefore cannot be overwritten or removed by the unprivileged (postgres) user.
+# This ensures the following:
+# - libraries and supporting files for extensions that are part of this Docker Image are immutable
+# - new libraries/files can be added to the container and updated by the postgres user
+ARG ALLOW_ADDING_EXTENSIONS=true
+RUN if [ "${ALLOW_ADDING_EXTENSIONS}" = "true" ]; then \
+        for pg in ${PG_VERSIONS}; do \
+            for dir in "$(/usr/lib/postgresql/${pg}/bin/pg_config --sharedir)/extension" "$(/usr/lib/postgresql/${pg}/bin/pg_config --pkglibdir)" "$(/usr/lib/postgresql/${pg}/bin/pg_config --bindir)"; do \
+                install --directory "${dir}" --group postgres --mode 1775 ; \
+            done; \
+        done ; \
+    fi
+
 ## Cleanup
 RUN apt-get remove -y ${BUILD_PACKAGES}
 RUN apt-get autoremove -y \
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 These are changes that will probably be included in the next release.
 
 ### Added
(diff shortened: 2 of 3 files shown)


itroot (approved):

Dockerfile
```diff
@@ -205,6 +205,22 @@ RUN if [ ! -z "${PG_LOGERRORS}" ]; then \
             && make clean && PG_CONFIG=/usr/lib/postgresql/${pg}/bin/pg_config make install || exit 1 ; \
         done; \
     fi
+
+# Allow Adding Extensions allows *new* files to be created, so that extensions can be added to a running container.
+# Existing files are still owned by root and have their sticky bit (the 1 in the 1750 permission mode) set,
+# and therefore cannot be overwritten or removed by the unprivileged (postgres) user.
+# This ensures the following:
+# - libraries and supporting files for extensions that are part of this Docker Image are immutable
```
alexeyklyukin, line 213: Clever approach. I wonder whether it would fit our goal of upgrading some extension (i.e. timescale) in place: the problem is the control file that has the same name; it will be owned by root and therefore won't allow to be overwritten from inside the container.

I guess we can mitigate that by setting timescale extension ownership to the postgres user.

Dockerfile
```diff
@@ -205,6 +205,22 @@ RUN if [ ! -z "${PG_LOGERRORS}" ]; then \
...
+# - new libraries/files can be added to the container and updated by the postgres user
+ARG ALLOW_ADDING_EXTENSIONS=
+RUN if [ "${ALLOW_ADDING_EXTENSIONS}" = "true" ]; then \
+        for pg in ${PG_VERSIONS}; do \
+            for dir in "$(/usr/lib/postgresql/${pg}/bin/pg_config --sharedir)/extension" "$(/usr/lib/postgresql/${pg}/bin/pg_config --pkglibdir)"; do \
```
alexeyklyukin, line 218: I wonder if we need to add bindir here as well; some extensions, i.e. pg_repack, ship binaries in addition to the libraries and sql files.

alexeyklyukin (approved):
Nice hack!

Dockerfile
feikesteenbergen: Yeah, I wanted to start with the most restrictions, relaxing them if needed, instead of adding too many privileges from the start

Dockerfile
feikesteenbergen: I'd be happy to add those once we run into those issues, although for binaries I think other mechanisms may be better.

Dockerfile
```diff
@@ -205,6 +205,22 @@ RUN if [ ! -z "${PG_LOGERRORS}" ]; then \
...
     fi
+
+# Allow Adding Extensions allows *new* files to be created, so that extensions can be added to a running container.
+# Existing files are still owned by root and have their sticky bit (the 1 in the 1750 permission mode) set,
+# and therefore cannot be overwritten or removed by the unprivileged (postgres) user.
+# This ensures the following:
+# - libraries and supporting files for extensions that are part of this Docker Image are immutable
+# - new libraries/files can be added to the container and updated by the postgres user
+ARG ALLOW_ADDING_EXTENSIONS=
+RUN if [ "${ALLOW_ADDING_EXTENSIONS}" = "true" ]; then \
```
feikesteenbergen, line 216: Wondering if we should just do it unconditionally

sgichohi (approved):

PR #102, opened 2021-04-07
PostgreSQL 13 - Soft deprecation of 11 images

feikesteenbergen:
# Support for PG 13 - prepare deprecation of PG 11

With this commit in, we can build Docker Images containing all the
software we need for PostgreSQL 13.

To make this work, we needed to fix a few things:

* `pg_auth_mon`: Bump tag to a PostgreSQL 13 supported version
* `timescale_analytics`: Only build PostgreSQL <version> and be explicit about
    the version when using `cargo pgx`, to avoid rust compilation
    errors.
* `timescaledb`: Ensure we only build 2.1+ for PostgreSQL 13

PostgreSQL 11 is not yet deprecated with this commit in, however both
the Makefile and GitHub Actions no longer build PostgreSQL 11 images.

### DIFF
--- a/.github/workflows/publish_image.yaml
+++ b/.github/workflows/publish_image.yaml
@@ -17,20 +17,20 @@ jobs:
     strategy:
       matrix:
         include:
-        # This is the 'default' image, it contains PostgreSQL 11 and 12, and defaults to PostgreSQL 12 binaries
+        # This is the 'default' image, it contains PostgreSQL 12 and 13, and defaults to PostgreSQL 13 binaries
+        - pg_major: "13"
+          pg_versions: "13 12"
+        # This is the 12 image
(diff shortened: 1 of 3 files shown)


alexeyklyukin:
With those changes, what would be the procedure of issuing v11 minor version upgrade?

feikesteenbergen:
> With those changes, what would be the procedure of issuing v11 minor version upgrade?

There would currently be no such thing anymore. We ourselves (timescale) don't use the v11 images anymore.
If someone needs PostgreSQL 11, they could use the PostgreSQL 12 image, but set the path to point to 11, this is because:

12 image contains 12 and 11
13 image contains 13 and 12

alexeyklyukin (approved):

Issue #101, opened 2021-03-31
Include pg_repack


feikesteenbergen:
Reported here: https://github.com/timescale/timescaledb-kubernetes/issues/274

PR #103, opened 2021-04-07
Include pg_repack and hypopg

feikesteenbergen:
In passing, remove a previous bugfix for PostGIS installation.
Fixes issue #101


alexeyklyukin (approved):
Nice one, although I'd note pg_repack requires a superuser to run.

PR #112, opened 2021-05-18
Include gdb and gdbserver

feikesteenbergen:
Having these tools available in the Image themselves simplifies many
debugging situations, whereas the cost of adding these to the Image is
measured in kB's.


alexeyklyukin (approved):
Yes!

PR #113, opened 2021-05-18
Add pg_stat_kcache libraries

feikesteenbergen:
https://github.com/powa-team/pg_stat_kcache is an extension that can
keep track of many system level metrics for us. It requires to be added
to the `shared_preload_libraries` to be useful.


sgichohi (approved):
nice

PR #123, opened 2021-06-21
Feike/build as postgres

feikesteenbergen:
# Install all compiled extensions as user postgres

This sets the ownership of the files to the postgres user, which allows
us to do hot-patching of these files in running containers.

Previously, we only allowed *adding* new files to the relevant
PostgreSQL directories, however there is the need to replace some files
in some instances, for example:

- replacing `timescaledb.so`
- replacing `timescaledb.control`

As long as these files were owned by root, there was no way we could
replace these files in an unprivileged container.

The root user still retains ownership of all PostgreSQL core and
Packaged extensions.

# Remove Sqlite FDW

We envisioned a use case for SQLite, however that use case did not
materialize, therefore we remove it from the image again.


itroot (approved):

alexeyklyukin (approved):
LGTM

PR #124, opened 2021-06-21
Change promscale extension to use PGX

cevian:
- [x] ~~do not merge: docker image will be tested by 1 customer after which it may or may not be merged~~

The promscale extension now uses the same PGX framework as
timescale-analytics. Update the docker file to be able to build
that.


feikesteenbergen (approved):

feikesteenbergen (approved):

cevian:
@feikesteenbergen @sgichohi I updated this PR with the new tagged version 0.2.0, which can now be deployed widely. So I think this is megeable now.

PR #129, opened 2021-06-30
Rename Analytics to Toolkit and up to 1.0.0

JLockerman:
This commit updates timescale-analytics to 1.0.0, which changes it's name to `timescaledb-toolkit`. This commit changes the relevant variables and extension name, but does not change the repo path, as that is not yet updated.


sgichohi:
Hey @JLockerman should we have both extension names in the image for backward compatibility?  A subsequent image could remove the analytics reference.

feikesteenbergen:
@sgichohi , yes, I think we should keep both, to allow a smooth transition, we should always be compatible with a previous Docker Image - at least for a couple of months.

scripts/version_info.sql
```diff
@@ -9,7 +9,7 @@ WITH versions(name, version) AS (
     FROM
         pg_available_extensions
     WHERE
-        name IN ('timescaledb', 'postgis', 'pg_prometheus', 'timescale_prometheus_extra', 'promscale', 'timescale_analytics')
+        name IN ('timescaledb', 'postgis', 'pg_prometheus', 'timescale_prometheus_extra', 'promscale', 'timescaledb_toolkit')
```
feikesteenbergen, line 12: nit, does not matter that much, as I don't think anyone relies on `timescale_analytics` appearing in the docker image labels, but is cheap enough to keep in here:
```suggestion
        name IN ('timescaledb', 'postgis', 'pg_prometheus', 'timescale_prometheus_extra', 'promscale', 'timescaledb_toolkit', 'timescale_analytics')
```

feikesteenbergen (approved):

sgichohi (approved):

PR #133, opened 2021-07-14
Prevent annoying message about missing cargo file

feikesteenbergen:


itroot (approved):
LGTM. Another way is probably to compile it in different stage of the build.

sgichohi (approved):

PR #150, opened 2021-08-24
Fix Rust Version

feikesteenbergen:
Before this commit, the rust installation that would build all rust
components of this Docker against the latest stable compiler of Rust.

A newer stable rust compiler may (and has previously) cause
compiler/test/clippy warnings that may fail a build of one of our Rust
components altogether.

As Rust releases are out of our control, it seems wise to fix the Rust
version to a specific one, so we can explicitly move to a newer version
if required.


alexeyklyukin (approved):
Looks good!

PR #153, opened 2021-08-26
Try Hot Forge to install custom extensions

feikesteenbergen:
Reword: This is to speed up the build process of the Docker Images when
developing this Docker Image.


Dockerfile
```diff
@@ -44,10 +44,9 @@ RUN apt-get install -y gdb gdbserver
 # These packages allow for a better integration for some containers, for example
 # daemontools provides envdir, which is very convenient for passing backup
 # environment variables around.
-RUN apt-get update && apt-get install -y dumb-init daemontools
+RUN apt-get install -y dumb-init daemontools
```
niksajakovljevic, line 47: Shouldn't we run `apt-get update` at least once to get new packages? I see it's remove in two places and I am not sure if we call it from some other place

niksajakovljevic (approved):
LGTM

Dockerfile
feikesteenbergen: We call it on line 27

PR #158, opened 2021-09-08
Use Ubuntu as a base for our Docker Image

feikesteenbergen:
- [ ] Don't merge quickly, need some broad discussion first

# Use Ubuntu as a base for our Docker Image

To allow us to use specific glibc 2.33+ features, we need to find a way
to run glibc 2.33. Running multiple glibc versions inside the same
container is something we'd like to avoid, we've seen multiple glibc
related bugs in our lifetime, adding multiple glibc versions in the mix
would make debugging harder.

Debian (and rust:debian) has served us well, however even Debian's
latest release (bullseye, August 2021) cannot give us glibc 2.33.

Ubuntu however does give us glibc 2.33 - as Ubuntu is based upon Debian
the changes required are not that big for this Docker Image. Most of the
tools we use will be the same across the board, as most of our tools our
installed using external repositories.

Some significant changes to get here are required though:

## install rust using rustup-init

There was some (valid) criticism[1], around using rust as a base for the
image, instead of Debian. As Rust does not offer an Ubuntu variant, we
are no longer using a rust-flavored Docker Image as a base, however, we
do copy the exact rust installation sequence - including comparing
hashes for sanity for installing Rust.

## deprecate `apt-key` usage

Ubuntu 21.04 seems to still allow apt-key, however, in the future
apt-key usage will be deprecated. As the tweaking to get this fixed for
Ubuntu 21.10 was already done, I decided it best to keep it in here.

## remove PostGIS 2.5 support

There is no packaged PostGIS 2.5 available for Ubuntu 21.04. We however
have seen no evidence of PostGIS 2.5 usage with our Docker Image, we do
see evidence of usage of PostGIS 3, which is the default. Dropping
support for 2.5 therefore does not seem to be such a major issue

## disable errors as warnings

These build errors are seen on building older TimescaleDB versions
against a recent glibc. For more details, see[2]

1: https://github.com/timescale/timescaledb-docker-ha/pull/140#discussion_r702882671
2: https://github.com/timescale/timescaledb/issues/2770

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
 ## The purpose of this Dockerfile is to build an image that contains:
 ## - timescale from (internal) sources
+## - many PostgreSQL extensions
 ## - patroni for High Availability
 ## - spilo to allow the github.com/zalando/postgres-operator to be compatible
 ## - pgBackRest to allow good backups
-
 
 # By including multiple versions of PostgreSQL we can use the same Docker image,
 # regardless of the major PostgreSQL Version. It also allow us to support (eventually)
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -17,37 +17,67 @@ ARG PG_MAJOR=13
 ## in relation to the total image size.
 ## By choosing a very basic base image, we do keep full control over every part
 ## of the build steps. This Dockerfile contains every piece of magic we want.
-FROM rust:1.54-slim-buster AS compiler
+FROM ubuntu:21.04 AS compiler
```
feikesteenbergen, line 20: This should really be 21.10, however 21.10 is currently utterly broken (it is in alpha, so expected)

Details: 21.10 fails with `apt-get update`:
```
#19 1.094 E: Problem executing scripts APT::Update::Post-Invoke 'rm -f /var/cache/apt/archives/*.deb /var/cache/apt/archives/partial/*.deb /var/cache/apt/*.bin || true'
#19 1.094 E: Sub-process returned an error code
------
error: failed to solve: executor failed running [/bin/sh -c apt-get update]: exit code: 100
make: *** [Makefile:128: prepare] Error 1
```

I'm not in the mood to fix alpha related problems currently, so using 21.04 to get the build passed would be good enough for now?

Dockerfile
```diff
@@ -70,14 +100,10 @@ RUN apt-get install -y python3-etcd python3-requests python3-pystache python3-ku
 
 # We install some build dependencies and mark the installed packages as auto-installed,
 # this will cause the cleanup to get rid of all of these packages
-ENV BUILD_PACKAGES="lsb-release git binutils libperl-dev libc6-dev patchutils gcc libc-dev make cmake libssl-dev python2-dev python3-dev devscripts equivs libkrb5-dev"
+ENV BUILD_PACKAGES="lsb-release wget git wget binutils libperl-dev gpg gpg-agent libc6-dev patchutils gcc libc-dev make cmake libssl-dev python2-dev python3-dev devscripts equivs libkrb5-dev"
```
feikesteenbergen, line 103: Double wget
```suggestion
ENV BUILD_PACKAGES="lsb-release wget git binutils libperl-dev gpg gpg-agent libc6-dev patchutils gcc libc-dev make cmake libssl-dev python2-dev python3-dev devscripts equivs libkrb5-dev"
```

Dockerfile
```diff
@@ -17,37 +17,67 @@ ARG PG_MAJOR=13
...
+RUN apt-get purge libappstream4
+
+RUN mkdir -p /build
+RUN chmod 777 /build
+WORKDIR /build/
 
+RUN wget -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | gpg --dearmor > /usr/share/keyrings/postgresql.keyring
 RUN for t in deb deb-src; do \
-        echo "$t http://apt.postgresql.org/pub/repos/apt/ $(lsb_release -s -c)-pgdg main" >> /etc/apt/sources.list.d/pgdg.list; \
+        echo "$t [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/postgresql.keyring] http://apt.postgresql.org/pub/repos/apt/ $(lsb_release -s -c)-pgdg main" >> /etc/apt/sources.list.d/pgdg.list; \
```
alexeyklyukin, line 40: 👍

alexeyklyukin (approved):
Looks great!

sgichohi (approved):
Wohoo.

Love the documentation as well.

itroot (approved):
:+1:

Dockerfile
```diff
@@ -17,37 +17,79 @@ ARG PG_MAJOR=13
...
 # create the postgres user as the first thing on our list
 RUN adduser --home /home/postgres --uid 1000 --disabled-password --gecos "" postgres
 
-# CI/CD may benefit a lot from using a specific package mirror
-ARG DEBIAN_REPO_MIRROR=""
-RUN echo 'APT::Install-Recommends "0";\nAPT::Install-Suggests "0";' > /etc/apt/apt.conf.d/01norecommend \
-    && if [ "${DEBIAN_REPO_MIRROR}" != "" ]; then \
-        sed -i "s{http://.*.debian.org{http://${DEBIAN_REPO_MIRROR}{g" /etc/apt/sources.list; \
-    fi
+RUN echo 'APT::Install-Recommends "0";\nAPT::Install-Suggests "0";' > /etc/apt/apt.conf.d/01norecommend
```
camscale, line 39: Usually you need to use `echo -e` to have it interpret `\n` - not sure why this works. The recommended portable way to do this is with printf. 

Also `APT::Install-Recommends` and `APT::Install-Suggests` are boolean options, so I think the string `"false"` reads better than `"0"`

Dockerfile
```diff
@@ -17,37 +17,79 @@ ARG PG_MAJOR=13
...
 
 # Install the highlest level dependencies
 RUN apt-get update \
-    && apt-get install -y curl ca-certificates locales gnupg1 lsb-release
+    && apt-get install -y curl ca-certificates gpg-agent wget gpg locales gnupg1 lsb-release
```
camscale, line 43: It is good to keep lists in alphabetical order. too often I've seen the same thing twice early and late in the list because its sometimes easy to miss elements when scanning.

Dockerfile
```diff
@@ -17,37 +17,79 @@ ARG PG_MAJOR=13
...
+
+RUN apt-get purge libappstream4
+
+RUN mkdir -p /build
+RUN chmod 777 /build
+WORKDIR /build/
 
+RUN wget -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | gpg --dearmor > /usr/share/keyrings/postgresql.keyring
```
camscale, line 51: If a program has an option to write output, it's better to use that than redirection as redirection can leave empty files around if the program errors. In this case, you could use `gpg --dearmor -o /usr/share/keyrings/postgresql.keyring`

(not relevant here, but this also works better with sudo).

Dockerfile
```diff
@@ -17,37 +17,79 @@ ARG PG_MAJOR=13
...
 # run inside the containers that use this Docker Image
-RUN apt-get update && apt-get install -y less jq strace procps
+RUN apt-get install -y less jq strace procps
 
 # For debugging it is very useful if the Docker Image contains gdb(server). Even though it is
 # not expected to be running gdb in a live instance often, it simplifies getting backtraces from
 # containers using this image
 RUN apt-get install -y gdb gdbserver
 
+# The next 2 run instructions lines are directly copied from https://github.com/rust-lang/docker-rust/blob/master/1.54.0/bullseye/Dockerfile
```
camscale, line 68: github links are best done with a commit hash instead of a branch, otherwise they break. And as I went to get a stable link, it seems this master link is broken already. (you can press `y` on a github page to turn a branch URL into a hash URL)

https://github.com/rust-lang/docker-rust/blob/21171fdd92e29acb045a41cd58b0d30d66aeaa7f/1.54.0/bullseye/Dockerfile is the stable link

Dockerfile
```diff
@@ -70,14 +112,10 @@ RUN apt-get install -y python3-etcd python3-requests python3-pystache python3-ku
 
 # We install some build dependencies and mark the installed packages as auto-installed,
 # this will cause the cleanup to get rid of all of these packages
-ENV BUILD_PACKAGES="lsb-release git binutils libperl-dev libc6-dev patchutils gcc libc-dev make cmake libssl-dev python2-dev python3-dev devscripts equivs libkrb5-dev"
+ENV BUILD_PACKAGES="lsb-release git wget binutils libperl-dev gpg gpg-agent libc6-dev patchutils gcc libc-dev make cmake libssl-dev python2-dev python3-dev devscripts equivs libkrb5-dev"
```
camscale, line 115: ditto alpha order

Dockerfile
```diff
@@ -116,8 +154,9 @@ FROM compiler as builder
 ARG PG_VERSIONS
 
 # timescaledb-tune, as well as timescaledb-parallel-copy
-RUN echo "deb https://packagecloud.io/timescale/timescaledb/debian/ $(lsb_release -s -c) main" > /etc/apt/sources.list.d/timescaledb.list
-RUN curl -L -s -o - https://packagecloud.io/timescale/timescaledb/gpgkey | apt-key add -
+RUN wget -O - https://packagecloud.io/timescale/timescaledb/gpgkey | gpg --dearmor > /usr/share/keyrings/timescaledb.keyring
```
camscale, line 157: ditto gpg -o

Dockerfile
```diff
@@ -194,8 +235,8 @@ RUN TS_VERSIONS="1.6.0 1.6.1 1.7.0 1.7.1 1.7.2 1.7.3 1.7.4 1.7.5 2.0.0-rc3 2.0.0
             && cd /build/timescaledb && git reset HEAD --hard && git clean -f -d -x && git checkout ${ts} \
             && rm -rf build \
             && if [ "${ts}" = "2.2.0" ]; then sed -i 's/RelWithDebugInfo/RelWithDebInfo/g' CMakeLists.txt; fi \
-            && PATH="/usr/lib/postgresql/${pg}/bin:${PATH}" ./bootstrap -DTAP_CHECKS=OFF -DCMAKE_BUILD_TYPE=RelWithDebInfo -DREGRESS_CHECKS=OFF -DGENERATE_DOWNGRADE_SCRIPT=ON -DPROJECT_INSTALL_METHOD="${INSTALL_METHOD}"${OSS_ONLY} \
-            && cd build && make -j 6 install || exit 1; \
+            && PATH="/usr/lib/postgresql/${pg}/bin:${PATH}" ./bootstrap -DTAP_CHECKS=OFF -DWARNINGS_AS_ERRORS=off -DCMAKE_BUILD_TYPE=RelWithDebInfo -DREGRESS_CHECKS=OFF -DGENERATE_DOWNGRADE_SCRIPT=ON -DPROJECT_INSTALL_METHOD="${INSTALL_METHOD}"${OSS_ONLY} \
+            && cd build && make install || exit 1; \
```
camscale, line 239: is `|| exit 1` needed? it just turns any non-zero exit code into 1, which is still an error. unless something relies on this being 1, I'd let it exit with the actual code it produced.

Also, instead of `cd build && make install`, you can just do `make -C build install`

feikesteenbergen: The `|| exit1` is to ensure that the `RUN` instructions ends right there; we could possibly do a `set -e` at the top though to make this shell terminate on any error.

As this is inside a for loop (over major postgresql versions), we cannot just `exit 0` here on success, we need to throw an error.

feikesteenbergen:
@camscale I'll take a few of your suggestions (exit 1) and do that in a separate commit/PR, as they don't really have to do with the switch to Ubuntu, but are valuable to do regardless.

PR #161, opened 2021-09-16
Release version 1.0.0

feikesteenbergen:
This is a major release, as the base Docker Image has changed from Debian to Ubuntu.
Our tests have not yet shown any issues with this update. We would advice anyone
that consumes these images to test that the new images also work for them in their
environment.

As this Docker Image has been in production for a while, it seems awkward to still not
be on version 1.0.0+, therefore, we mark this occasion with releasing version 1.0.0.


alexeyklyukin (approved):
LGTM

PR #183, opened 2021-11-23
PostgreSQL 14 support

feikesteenbergen:
This PR results in docker images being built for PostgreSQL 14.

# Default to building images with PostgreSQL 14 and 13

# Use latest tagged protected_roles
    
Version 1.0.0 can be compiled against PostgreSQL 14

# Separate complex RUN steps into scripts
    
The steps to install timescaledb and timescaledb-toolkit have become
quite complex. To simplify their maintenance, we move the commands into
their own scripts, so they can be checked, inspected and reviewed in
their own context.

The logic for version compatibility has changed slightly, to support
PostreSQL 14 as well as to simplify the code.


.github/workflows/publish_image.yaml
```diff
@@ -18,10 +18,18 @@ jobs:
     strategy:
       matrix:
         include:
-        # This is the 'default' image, it contains PostgreSQL 12 and 13, and defaults to PostgreSQL 13 binaries
+        # This is the 'default' image, it contains PostgreSQL 14 and 12, and defaults to PostgreSQL 14 binaries
```
feikesteenbergen, line 21:
```suggestion
        # This is the 'default' image, it contains PostgreSQL 14 and 13, and defaults to PostgreSQL 14 binaries
```

build_scripts/install_timescaledb.sh
```diff
@@ -0,0 +1,50 @@
...
+is_supported() {
+    MINOR="$(echo "$1" | cut -d. -f2)"
+    if [ "${PGVERSION}" -lt 12 ] && [ "${MINOR}" -ge 4 ]; then return 1; fi
+    if [ "${PGVERSION}" -ge 14 ] && [ "${MINOR}" -lt 5 ]; then return 1; fi
+    if [ "${PGVERSION}" -ge 13 ] && [ "${MINOR}" -lt 1 ]; then return 1; fi
+    return 0
+}
+
+for TAG in "$@"; do
+    cd "${PWD}"
```
feikesteenbergen, line 26:
```suggestion
```

Dockerfile
```diff
@@ -8,7 +8,7 @@
 # By including multiple versions of PostgreSQL we can use the same Docker image,
 # regardless of the major PostgreSQL Version. It also allow us to support (eventually)
 # pg_upgrade from 12 to 13, so we need all the postgres & timescale libraries for all versions
```
feikesteenbergen, line 10:
```suggestion
# pg_upgrade from one major version to another, so we need all the postgres & timescale libraries for all versions
```

itroot (approved):

PR #185, opened 2021-12-02
Feike/ts251

feikesteenbergen:


itroot (approved):
:+1:

CHANGELOG.md
```diff
@@ -16,6 +16,20 @@ These are changes that will probably be included in the next release.
 
 ### Fixed
 
+## [v1.1.0] - 2021-12-02
+
+### Removed
+ * No longer build PostgreSQL 12 Docker Images in CI/CD. PostgreSQL 12 binaries are included
```
alexeyklyukin, line 22: That sentence is slightly confusing, there are 3 different postgres versions mentioned. We default to building v13 and v14, and we seem to [build v12 oss](https://github.com/timescale/timescaledb-docker-ha/blob/f12aab0a4f898dc64daf95960df5299e789fd18e/.github/workflows/publish_image.yaml#L30).

Does it mean we no longer update v12?

feikesteenbergen: I'll reword (after CI/CD has given the all clear as to not wait another hour).

Every Docker Image we release is tagged with a major PostgreSQL version.
Such an image will contain:

- the major PostgreSQL version of the tag
- the major PostgreSQL version - 1

That means that as of this release, we no longer build images containing PostgreSQL 12 and 11.
We do however build images containing PostgreSQL 13 and 12.

Therefore, PostgreSQL 11 no longer gets patches, PostgreSQL 12 *does* if you use the PostgreSQL 13 image.


> and we seem to build v12 oss.

We build the v13 oss, which includes pg13 and pg12.

This is quite confusing, so I'm thinking on how to word this in such a way that clarity prevails!

PR #189, opened 2021-12-16
Add pg_cron to shared_preload_libraries

jflambert:
pg_cron was installed but not available.

Users still need to run `CREATE EXTENSION pg_cron;`


feikesteenbergen:
I don't think we should add any libraries to `shared_preload_libraries` by default. `timescaledb` being the exception, as the whole point of this Docker Image is to run `timescaledb`.

PR #198, opened 2022-01-28
Sanity-check database objects before installing timescale extension

alexeyklyukin:
Make sure there are no pre-existing schemas for "create extension" and
no triggers defined on timescale schemas for "alter extension update".

### DIFF
--- a/scripts/timescaledb/before-create.sql
+++ b/scripts/timescaledb/before-create.sql
@@ -3,18 +3,19 @@
 DO $$
 BEGIN
     IF EXISTS(
-    SELECT
-        1
-    FROM
-        pg_catalog.pg_namespace
-            WHERE nspname IN (
-                '_timescaledb_catalog',
-                '_timescaledb_internal'
-                '_timescaledb_cache',
-                '_timescaledb_config',
-                'timescaledb_experimental',
-                'timescaledb_information'
-            )
+        SELECT
+            1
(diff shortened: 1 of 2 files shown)


scripts/timescaledb/before-create.sql
```diff
@@ -0,0 +1,24 @@
...
+-- Note: this list assumes any new schemas added to the extension should be present here as well.
+DO $$
+BEGIN
+    IF EXISTS(
+    SELECT
+        1
+    FROM
+        pg_catalog.pg_namespace
+            WHERE nspname IN (
+                '_timescaledb_catalog',
```
alexeyklyukin, line 11: or should we "take over" _timescaledb_% instead of hard-coding schemas in this list?

scripts/timescaledb/before-create.sql
```diff
@@ -0,0 +1,24 @@
+-- Check if one of the timescaledb schemas are present in the database, while the extension is missing
+-- Note: this list assumes any new schemas added to the extension should be present here as well.
+DO $$
+BEGIN
+    IF EXISTS(
```
alexeyklyukin, line 5: Actually, since this is called before the CREATE EXTENSION command, need to check that timescaledb is not yet installed.

alexeyklyukin:
Closing for now, need to revisit the approach.

PR #200, opened 2022-02-02
Update toolkit to version 1.5.0

WireBaron:


feikesteenbergen (approved):
Perhaps we can drop a few older versions at some point to reduce build times?

feikesteenbergen:
> error: pathspec 'forge-stable-1.5.0' did not match any file(s) known to git

There is no `forge-stable-1.5.0` tag, in order for this to build, the tag requested to be built needs to be pushed to the toolkit repository

@WireBaron

feikesteenbergen (changes requested):
The tag/branch should exist

feikesteenbergen:
Closing in favor of immediately providing 1.5.1

PR #210, opened 2022-02-17
Stamp v1.1.7

feikesteenbergen:
- [x] do some smoketests, we can use the image built here: https://github.com/timescale/timescaledb-docker-ha/runs/5228954016?check_suite_focus=true


alexeyklyukin (approved):
Awesome!

PR #216, opened 2022-03-08
Use Ubuntu 22.04 LTS as a base for our Docker Image

feikesteenbergen:
- [x] Wait for Ubuntu Release 22.04 to be published, expected on April 21, 2022

The https://ubuntu.com/about/release-cycle informs us that support for
Ubuntu 21.04 has ended, and we should therefore upgrade to a newer
Ubuntu release.

As Ubuntu 21.10 is also an interim release (supported for 9 months only)
and 22.04 is around the corner which is a LTS release with 5 years worth
of support it seems best to switch directly to 22.04


sgichohi (approved):

JamesGuthrie:
🎉

PR #218, opened 2022-03-10
Use Ubuntu 21.10 as a base for our Docker Image

feikesteenbergen:
The https://ubuntu.com/about/release-cycle informs us that support for
Ubuntu 21.04 has ended, and we should therefore upgrade to a newer
Ubuntu release.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,8 +31,9 @@ ARG PG_MAJOR=13
 ## the changes required are not that big for this Docker Image. Most of the
 ## tools we use will be the same across the board, as most of our tools our
 ## installed using external repositories.
-FROM ubuntu:21.04 AS compiler
+FROM ubuntu:21.10 AS compiler
 
+ENV DEBIAN_FRONTEND=noninteractive
 # We need full control over the running user, including the UID, therefore we
 # create the postgres user as the first thing on our list
 RUN adduser --home /home/postgres --uid 1000 --disabled-password --gecos "" postgres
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -155,8 +156,10 @@ FROM compiler as builder
 ARG PG_VERSIONS
 
 # timescaledb-tune, as well as timescaledb-parallel-copy
+# TODO: Replace `focal` with `$(lsb_release -s -c)` once packages are available
+# for Ubuntu 21.10
 RUN wget -O - https://packagecloud.io/timescale/timescaledb/gpgkey | gpg --dearmor --output /usr/share/keyrings/timescaledb.keyring
-RUN echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/timescaledb.keyring] https://packagecloud.io/timescale/timescaledb/ubuntu/ $(lsb_release -s -c) main" > /etc/apt/sources.list.d/timescaledb.list
+RUN echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/timescaledb.keyring] https://packagecloud.io/timescale/timescaledb/ubuntu/ focal main" > /etc/apt/sources.list.d/timescaledb.list
```
alexeyklyukin, line 162: The former line seems more universal, why did we change it?

feikesteenbergen: Because there are no packages published for the Ubuntu codename `impish`, yet the `focal` packages are good enough, as we only install statically compiled binaries using this repository.

Once we switch to Ubuntu 22.04, we should be able to switch it back to the previous line, see the TODO added slightly above.

alexeyklyukin: Yeah, sorry, read the comment, but couldn't resolve the conversation, as the action failed.

alexeyklyukin (approved):
Nice!

PR #221, opened 2022-03-21
Build TimescaleDB Toolkit from public repo again

feikesteenbergen:
This is basically a revert of commit 922657d6, which was needed to
include a security fix that wasn't part of the public repository yet.


alexeyklyukin (approved):
Looks good!

PR #224, opened 2022-04-05
Add Patroni updates for new static_primary config option

thedodd:
- Related to https://github.com/timescale/patroni/pull/1

### DIFF
--- a/.github/workflows/publish_image.yaml
+++ b/.github/workflows/publish_image.yaml
@@ -14,6 +14,7 @@ env:
   TIMESCALE_TSDB_ADMIN: "1.0.0"
   TIMESCALEDB_TOOLKIT_REPO: github.com/timescale/timescaledb-toolkit
   TIMESCALEDB_TOOLKIT_EXTENSIONS: forge-stable-1.3.1 1.5.1-cloud 1.6.0
+  TIMESCALE_STATIC_PRIMARY: ''
 jobs:
   publish-image:
     name: Publish the Docker Images
@@ -23,6 +24,11 @@ jobs:
         # This is the 'default' image, it contains PostgreSQL 14 13 and 12, and defaults to PostgreSQL 14 binaries
         - pg_major: "14"
           pg_versions: "14 13 12"
+        # This is the 'default' image, along with Patroni changes to support static primaries
+        - pg_major: "14"
+          pg_versions: "14"
+          patroni_static_primary: "true"
+          docker_tag_postfix: "-patroni-static-primary"
         # This is the PostgreSQL 14 image containing only oss software
         - pg_major: "14"
           pg_versions: "14 13 12"
@@ -75,6 +81,7 @@ jobs:
           DOCKER_ECR_URL: ${{ secrets.AWS_ACCOUNT_ID }}.dkr.ecr.us-east-1.amazonaws.com/timescaledb-ha
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          TIMESCALE_STATIC_PRIMARY: ${{ matrix.patroni_static_primary }}
         # Note that our ECR registry has immutability guarantees, therefore we do not push
         # the mutable ('latest') tags to the ECR registry
         run: |
--- a/Dockerfile
+++ b/Dockerfile
@@ -146,7 +146,15 @@ RUN for postgisv in ${POSTGIS_VERSIONS}; do \
 # This need to be done after the PostgreSQL packages have been installed,
 # to ensure we have the preferred libpq installations etc.
 RUN apt-get install -y python3-etcd python3-requests python3-pystache python3-kubernetes python3-pysyncobj
-RUN apt-get install -y patroni
+RUN apt-get install -y patroni=2.1.3-\*
+# Patch Patroni code with changes from https://github.com/timescale/patroni/pull/1
+# NOTE: This is a temporary solution until changes land upstream.
+ARG TIMESCALE_STATIC_PRIMARY
+RUN if [ "${TIMESCALE_STATIC_PRIMARY}" != "" ]; then \
+    wget -qO- https://raw.githubusercontent.com/timescale/patroni/v2.2.0-beta.3/patroni/ha.py > /usr/lib/python3/dist-packages/patroni/ha.py && \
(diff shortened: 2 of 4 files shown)


alexeyklyukin (approved):
Looks good!

thedodd:
@alexeyklyukin yes sir, that is actually exactly what I did in this case. The code that is copied into the container is based on a tag which I've added to our internal Patroni fork `:)`. EG, the `v2.2.0-beta.0` in `https://raw.githubusercontent.com/timescale/patroni/v2.2.0-beta.0/patroni/ha.py` being pulled into the container.

feikesteenbergen:
@thedodd It seems it fails, as at the point you're doing the `wget` you are already the `postgres` user.

You should do this as the `root` user, so perhaps switch to the `root` user first and then switch back to `postgres`?

Dockerfile
```diff
@@ -453,6 +453,11 @@ RUN rm /etc/pgbackrest.conf && ln -s "${PGBACKREST_CONFIG}" /etc/pgbackrest.conf
 ## to make this work nicely, they need to exist though
 RUN for i in $(seq 0 7); do touch "${PGLOG}/postgresql-$i.log" "${PGLOG}/postgresql-$i.csv"; done
 
+## Patch Patroni code with changes from https://github.com/timescale/patroni/pull/1
+## This is a temporary solution until changes land upstream.
+RUN wget -qO- https://raw.githubusercontent.com/timescale/patroni/v2.2.0-beta.0/patroni/ha.py > /usr/lib/python3/dist-packages/patroni/ha.py && \
+    wget -qO- https://raw.githubusercontent.com/timescale/patroni/v2.2.0-beta.0/patroni/config.py > /usr/lib/python3/dist-packages/patroni/config.py
+
```
feikesteenbergen, line 460:
```suggestion
## Patch Patroni code with changes from https://github.com/timescale/patroni/pull/1
## This is a temporary solution until changes land upstream.
USER root
RUN wget -qO- https://raw.githubusercontent.com/timescale/patroni/v2.2.0-beta.0/patroni/ha.py > /usr/lib/python3/dist-packages/patroni/ha.py && \
    wget -qO- https://raw.githubusercontent.com/timescale/patroni/v2.2.0-beta.0/patroni/config.py > /usr/lib/python3/dist-packages/patroni/config.py
USER root
```

thedodd: Thanks boss. I actually decided to just move this code up a bit, so that it appears within the RUN clause which installs Patroni itself.

thedodd:
Ok, this has now been updated based on the `v2.2.0-beta.1` tag on the Patroni PR.

Dockerfile
```diff
@@ -146,7 +146,12 @@ RUN for postgisv in ${POSTGIS_VERSIONS}; do \
 # This need to be done after the PostgreSQL packages have been installed,
 # to ensure we have the preferred libpq installations etc.
 RUN apt-get install -y python3-etcd python3-requests python3-pystache python3-kubernetes python3-pysyncobj
-RUN apt-get install -y patroni
+RUN apt-get install -y patroni=2.1.3-\*
+# Patch Patroni code with changes from https://github.com/timescale/patroni/pull/1
+# NOTE: This is a temporary solution until changes land upstream.
+RUN wget -qO- https://raw.githubusercontent.com/timescale/patroni/v2.2.0-beta.1/patroni/ha.py > /usr/lib/python3/dist-packages/patroni/ha.py && \
```
alexeyklyukin, line 152: I think we should make this part optional (perhaps using docker variables) and keep this disabled by default for now.

thedodd: That's a good idea. We can add the param to a build matrix so that we can stamp-out a version with these changes, and a version without. Then if we ever need to switch due to bugs &c, we can do so.

thedodd: Ok, I think I've got a working solution. Testing now.

Dockerfile
```diff
@@ -146,7 +146,15 @@ RUN for postgisv in ${POSTGIS_VERSIONS}; do \
...
-RUN apt-get install -y patroni
+RUN apt-get install -y patroni=2.1.3-\*
+# Patch Patroni code with changes from https://github.com/timescale/patroni/pull/1
+# NOTE: This is a temporary solution until changes land upstream.
+ARG TIMESCALE_STATIC_PRIMARY
+RUN if [ "${TIMESCALE_STATIC_PRIMARY}" != "" ]; then \
+    wget -qO- https://raw.githubusercontent.com/timescale/patroni/v2.2.0-beta.1/patroni/ha.py > /usr/lib/python3/dist-packages/patroni/ha.py && \
+    wget -qO- https://raw.githubusercontent.com/timescale/patroni/v2.2.0-beta.1/patroni/config.py > /usr/lib/python3/dist-packages/patroni/config.py && \
+    wget -qO- https://raw.githubusercontent.com/timescale/patroni/v2.2.0-beta.1/patroni/validator.py > /usr/lib/python3/dist-packages/patroni/validator.py; \
+    fi
```
thedodd, line 157: Hey @feikesteenbergen && @alexeyklyukin just wanted to ping you guys here again.
- these bits are now optionally included via build-arg;
- a new build matrix stage has been added to the tagged CI pipeline, which will suffix the release tag with `-patroni-static-primary` so that we can easily differentiate the images;
- this will allow us to enable these changes throughout our stack, and easily rollback the Patroni changes with a minimal image change.

.github/workflows/publish_image.yaml
```diff
@@ -23,6 +24,11 @@ jobs:
         # This is the 'default' image, it contains PostgreSQL 14 13 and 12, and defaults to PostgreSQL 14 binaries
         - pg_major: "14"
           pg_versions: "14 13 12"
+        # This is the 'default' image, along with Patroni changes to support static primaries
+        - pg_major: "14"
+          pg_versions: "14 13 12"
```
feikesteenbergen, line 29: Nit, while we're evaluating this option, my guess is we'll only need PG14 binaries. That would reduce the image size and build time.

Unless we want to experiment with this feature on existing PG12 and PG13 deployments?
```suggestion
          pg_versions: "14 13 12"
```

feikesteenbergen (approved):
Looks good to me, especially as we now do not touch the default Docker Image being built!

.github/workflows/publish_image.yaml
thedodd: That's fine by me. I don't have any reason to target 12 & 13. Will remove those.

thedodd: Done.

feikesteenbergen (approved):

PR #228, opened 2022-04-19
Bump cloudutils to v1.1.3

inchuev:
Contains several bugfixes:
  * Role propagation commands now properly propagate NOINHERIT
  * Newly added data nodes will now correctly have all previosly created roles set


feikesteenbergen:
LGTM, build time is long, so we'll just have to wait and see what the CI/CD tells us.

PR #245, opened 2022-05-10
Install timescaledb_toolkit by default in databases

JamesGuthrie:


feikesteenbergen (changes requested):
For merging to master I think we should never use a commit hash, always a tag, even if that tag isn't production.

These are 2 separate commits doing separate things, can we split them in separate PR's?

I think the install timescaledb_toolkit by default is ok to be merged as is, and we can included it in our image ASAP.

JamesGuthrie:
> For merging to master I think we should never use a commit hash, always a tag, even if that tag isn't production.
> 
> These are 2 separate commits doing separate things, can we split them in separate PR's?
> 
> I think the install timescaledb_toolkit by default is ok to be merged as is, and we can included it in our image ASAP.

Agree entirely. The first commit's change was included as part of a separate PR which has already been merged. I had it in here to validate that the image would build. I will rebase.

JamesGuthrie:
@feikesteenbergen on another note, could you perhaps explain how/when `post_init.sh` is actually used/run? I modified it because I thought I might have to, but I don't understand how/when spilo is used.

feikesteenbergen:
I think `post_init.sh` is never run by us. I also don't think anyone is using this Docker Image as a `spilo` replacement anymore, but I could be wrong. I believe that integration broke multiple years ago.

I guess for now, it's fine that it is in there, as `spilo` (just as our Cloud) does not use the entrypoint init scripts to bootstrap the database.

Perhaps we should actually remove the spilo compatibility if it is broken - it only confuses.

feikesteenbergen (approved):

PR #253, opened 2022-05-20
Feike/patroni single primary multi version

feikesteenbergen:
For those environments that want to run the experimental Patroni version, but still support PostgreSQL 12 and 13 databases, we should ensure the final Docker Image contains those versions as well.


thedodd (approved):
+1

PR #261, opened 2022-05-31
Update Patroni with latest bug fixes (pre-release)

thedodd:


itroot:
I was using python quite a long time ago, but have a question - does the original package contains `pyc` files?

```
postgres@m3ky49vsii-an-0:~$ ls /usr/lib/python3/dist-packages/patroni/__pycache__/
__init__.cpython-39.pyc  async_executor.cpython-39.pyc  daemon.cpython-39.pyc      log.cpython-39.pyc              request.cpython-39.pyc    version.cpython-39.pyc
__main__.cpython-39.pyc  config.cpython-39.pyc          exceptions.cpython-39.pyc  psycopg.cpython-39.pyc          utils.cpython-39.pyc
api.cpython-39.pyc       ctl.cpython-39.pyc             ha.cpython-39.pyc          raft_controller.cpython-39.pyc  validator.cpython-39.pyc
postgres@m3ky49vsii-an-0:~$ 

```

Looks like it does... @thedodd I'm not sure it will be working. Could be wrong though.

itroot:
https://docs.python.org/3/tutorial/modules.html#compiled-python-files - I'm not sure how it is now in python3, but in python2 it could be that you end running code from `pyc` which changed `py` file.

thedodd:
The Python byte code (.pyc) is JIT compiled by the interpreter, so it is present after first use. The package itself does not include the byte code. You can verify this by looking at the [debian package manifests](https://packages.debian.org/bookworm/all/patroni/filelist).

feikesteenbergen (approved):

alexeyklyukin (approved):
Looks good, thank you!

itroot:
Folks, probably should be closed, as we had plans to roll patroni release?

alexeyklyukin:
@itroot I think we are still waiting for 2.1.4 packages on Ubuntu

thedodd:
Ok, I do think that this if finally ready! Pinging folks for another round of review.

feikesteenbergen:
LGMT. I think we can still get away with waiting a bit, as PG14.4 is about to be released today.

PR #268, opened 2022-06-16
Feike/revert dropping of older tsdb

feikesteenbergen:


alexeyklyukin (approved):
Pity we can't remove those yet.

PR #295, opened 2022-08-10
includes tds_fdw libraries

fagner-icarotech:


fagner-icarotech:
includes tds_fdw libraries

Dockerfile
```diff
@@ -98,7 +98,7 @@ RUN set -eux; \
 RUN apt-get update && apt-get install -y dumb-init daemontools
 
 RUN apt-get update \
-    && apt-get install -y postgresql-common pgbouncer pgbackrest lz4 libpq-dev libpq5 pgtop \
+    && apt-get install -y postgresql-common pgbouncer pgbackrest lz4 libpq-dev libpq5 pgtop libsybdb5 freetds-dev freetds-common gnupg gcc make \
```
feikesteenbergen, line 101: adding gnupg gcc make and all the others on this line isn't great.

On line 114 make, gmake, gcc and gpg are already installed, adding a separte `RUN` command *after* that line that would would IMHO be preferable over this. Somewhere after lines 127 seems more appropriate, because at that point many postgresql packages are already installed.

feikesteenbergen (changes requested):
I don't have a strong opinion of adding this. What benefit does it bring?
What are the costs of this feature:

- increase in image size
- increase in build times?

PR #297, opened 2022-08-11
pg 14.5 and 13.8

manute:
Based on this https://www.postgresql.org/developer/roadmap/, that will be released today


alexeyklyukin (approved):
Looks good!

PR #323, opened 2022-11-03
Include OSM extension

niconosenzo:


niconosenzo:
Test:
```sql
postgres=# SELECT * FROM pg_available_extensions where name = 'timescaledb_osm';
      name       | default_version | installed_version |           comment            
-----------------+-----------------+-------------------+------------------------------
 timescaledb_osm | 0.0.2           |                   | Manages object storage on S3
(1 row)

postgres=# CREATE EXTENSION timescaledb_osm;
CREATE EXTENSION
postgres=# \dx
                                                                List of installed extensions
        Name        | Version |   Schema   |                                                  Description                                                   
--------------------+---------+------------+----------------------------------------------------------------------------------------------------------------
 file_fdw           | 1.0     | public     | foreign-data wrapper for flat file access
 hotforge           | 0.1.39  | hotforge   | Hot Forge enables zero-downtime upgrades of PostgreSQL files inside a container.                              +
                    |         |            | This extension provides an interface to execute the hot-forge binary,                                         +
                    |         |            | as well as some Foreign Data Wrappers that allows one to query the patch catalog for the status of the patches
 pg_auth_mon        | 1.0     | public     | monitor connection attempts per user
 pg_stat_statements | 1.9     | public     | track planning and execution statistics of all SQL statements executed
 pgcrypto           | 1.3     | public     | cryptographic functions
 plpgsql            | 1.0     | pg_catalog | PL/pgSQL procedural language
 timescaledb        | 2.8.1   | pg_catalog | Enables scalable inserts and complex queries for time-series data
 timescaledb_osm    | 0.0.2   | pg_catalog | Manages object storage on S3
(8 rows)

postgres=#
```

.github/workflows/build_image.yaml
```diff
@@ -19,6 +19,7 @@ env:
   TIMESCALE_STATIC_PRIMARY: "true"
   TIMESCALE_TSDB_ADMIN: "1.0.0"
   TIMESCALEDB_TOOLKIT_EXTENSIONS: 1.6.0 1.7.0 1.8.0 1.10.1 1.11.0
+  TIMESCALE_OSM_EXTENSION: "0.0.2"
```
sgichohi, line 22: should we add 0.0.2 and 0.0.3 as a list?

sgichohi (approved):
Nice! nit on 0.0.3

.github/workflows/build_image.yaml
niconosenzo: @sgichohi I changed it for installing every osm:* release for each PG version, do you think it worth to separate this as an external script (following toolkit pattern) or it's fine how it is now ?
 
https://github.com/timescale/timescaledb-docker-ha/blob/de5fe6c846229e4d93acdd5a1091b052b1b5ba43/Dockerfile#L303-L323

sgichohi: cc @feikesteenbergen might know best.

feikesteenbergen (changes requested):
This requires an additional 30 minutes to build if I skim through the logs.
Compiling pgx and the osm extension multiple times for every major PostgreSQL version should really not be done in the Docker Image. We don't wont to go back to hours of build times for this image.

Have a look at how promscale does it for example, it publishes compiled artifacts.

https://github.com/timescale/timescaledb-docker-ha/blob/master/build_scripts/install_promscale.sh

feikesteenbergen (approved):

PR #341, opened 2022-12-26
Minor shell test cleanup

graveland:
Just some minor tweaks to make the shell tests more canonical, and to drop cloning with credentials
for the timescaledb repository, since it's public anyway.


graveland:
These changes are getting included in the split repo stuff

PR #346, opened 2023-01-11
Use Patroni dcs-failsafe-mode branch

andrewcharlton:
We currently override the default `patroni` image with our own fork, with a `static-primary` patch applied to prevent patroni demoting postgres to read-only in the case of patroni being unable to reach etcd.  However, this only works for the single replica case.

There is an upstream PR (https://github.com/zalando/patroni/pull/2379), which generalises this to the HA case. With this patch, if patroni is unable to contact etcd for whatever reason, it will only demote the primary if it is unable to reach the other patroni instances. This way, if it only the control plane that is down, we can still service writes.

I have copied the PR branch to our fork, this PR just switches to using it instead of our existing static-primary branch.


feikesteenbergen:
LGTM. In a few weeks, we should be able to follow upstream again (using debian packages!) and therefore remove all the dependencies on our explicit fork.

feikesteenbergen (approved):

andrewcharlton:
> In a few weeks, we should be able to follow upstream again (using debian packages!) and therefore remove all the dependencies on our explicit fork.

That's the dream!

PR #347, opened 2023-01-17
Update OSM ext versions

niconosenzo:


.github/workflows/build_image.yaml
```diff
@@ -19,7 +19,7 @@ env:
   TIMESCALE_STATIC_PRIMARY: "true"
   TIMESCALE_TSDB_ADMIN: "1.0.0"
   TIMESCALEDB_TOOLKIT_EXTENSIONS: 1.6.0 1.7.0 1.8.0 1.10.1 1.11.0 1.12.0 1.12.1 1.13.0
-  TIMESCALE_OSM_EXTENSION: 0.0.2 0.0.3
+  TIMESCALE_OSM_EXTENSION: 0.0.4 0.0.5
```
graveland, line 22:
```suggestion
  TIMESCALE_OSM_EXTENSION: 0.0.5
```

The OSM extension doesn't install versioned files - it ends up as `/usr/lib/postgresql/<version>/lib/timescaledb_osm.so`, so there's no need to have multiple versions here.

graveland (changes requested):
Just a minor change

.github/workflows/build_image.yaml
niconosenzo: @graveland thanks, updated.

graveland (approved):
lgtm

.github/workflows/build_image.yaml
```diff
@@ -19,7 +19,7 @@ env:
   TIMESCALE_STATIC_PRIMARY: "true"
   TIMESCALE_TSDB_ADMIN: "1.0.0"
   TIMESCALEDB_TOOLKIT_EXTENSIONS: 1.6.0 1.7.0 1.8.0 1.10.1 1.11.0 1.12.0 1.12.1 1.13.0
-  TIMESCALE_OSM_EXTENSION: 0.0.2 0.0.3
+  TIMESCALE_OSM_EXTENSION: "0.0.5"
```
gayyappan, line 22: do we need 0.0.4 as well?

graveland: You'd have to modify the OSM extension to install as versioned files, as currently, each one overwrites the previous one. It's not useful right now to have more than one. ie., `/usr/lib/postgresql/14/lib/timescaledb_osm.so` is where each version of OSM is installed if you're building 0.0.4 + 0.0.5.

gayyappan: Thanks for the clarification.  
Let's say we modify the process and create versioned .so i.e. timescaledb-osm-0.0.4.so and timescaledb-osm-0.0.5.so. 
- What would change during deployment?  
- What are the missing steps between the current install process and a world with versioned .so? (Is it as simple as just creating a link from timescaledb_osm.so TO timescaledb_osm_<latest-version>.so

PR #349, opened 2023-01-19
allow build-time specification of UID/GID

fsiler:
This issue comes up when you're trying to use a volume map and the host doesn't happen to have postgres:postgres on 1000:1000.  Using this patch you can specify an alternate build-time UID and GID.

Also fixes the fact that `master` was renamed to `main`.


graveland:
#350 addresses this in a different way- would that PR solve your issue?

fsiler:
yes, #350 is a much more generic and clean solution- I'm closing this PR!

Issue #311, opened 2022-09-19
deployment as in docker stack has data volume permission problems


pezzacolori:
I need Timescale together with postgis.

By deploying  timescale/timescaledb-ha:pg14-latest  
in a docker stack  

  postgres:
    ports:
      - 54xx:5432
    image: timescale/timescaledb-ha:pg14-latest
    volumes:
      - pg_data:/var/lib/postgresql/data
    environment:
      - POSTGRES_PASSWORD=xxxx
      - PGDATA=/var/lib/postgresql/data/pgdata
    networks:
      - box-net

it raises an error:

mkdir: cannot create directory ‘/var/lib/postgresql/data/pgdata’: Permission denied

On the contrary using the     timescale/timescaledb:latest-pg14  
image there are no problems and the database is created automatically.

Maybe it is a problem linked to the packaged linux version (apline in plain timescaledb, ubuntu in timescaledb-ha) ?
Do you know how to solve the issue?

Cheers,
Boris

VMM-Mtech:
See #272 and #290. Is your case that you have created the volume with the standard image, and then mounting it to the ha container? Then the root cause seems to be the fact that standard image uses uid:gid 70:70, and the ha container uses 1000:1000.

graveland:
Let me know if the latest image works for you now!

My testing with ```--user 1234:1111 -v `pwd`/db:/home/pgdata -ePGDATA=/home/pgdata <image>```, stopping, and restarting with different --user values created a database, and then successfully restarted. Hopefully this will help?

The postgresql user is still always `postgres`, but nss-wrapper tricks postgresql into thinking ownership is correct. You should have an easier time using this image and just modifying `PGDATA` and `--user`. Specifically try to avoid using locations in the default `postgres` user home directory `/home/postgres` when you're using either a different user, or a different `PGDATA` value.

samul-1:
I'm running into the same issue. Here's how I've defined the service in my docker-compose:

```
  timescaledb:
    image: timescale/timescaledb-ha:pg16
    volumes:
      - ./timescaledb_data:/home/postgres/pgdata/data/
    environment:
      POSTGRES_USER: puser
      POSTGRES_PASSWORD: ppass
    expose:
      - "5432"
    ports:
      - "5433:5432"
    extra_hosts:
      - "host.docker.internal:host-gateway"
```

Here's what I get when running the container:

```
timescaledb-1  | using nss-wrapper
[...]

Issue #290, opened 2022-07-26
Databases are deleted after image update


CM000n:
Hello all,
First of all, thank you very much for your work.
I switched from the official Timescale Docker image with Postgis to your HA Docker image some time ago.

My setup looks like this: I have pinned the timescale version to 14, and a Watchtower container ensures a regular update.
Here is an excerpt of my Docker compose.yaml

```yaml
  timescale:
    container_name: timescale
    restart: always
    image: timescale/timescaledb-ha:pg14-latest
    networks:
      - proxyv6
    volumes:
      - /srv/dev-disk-by-uuid-someuuid/data/docker_data/timescale_ha:/var/lib/postgresql/data
      - /etc/localtime:/etc/localtime:ro
    environment:
      - POSTGRES_USER=MYUSER
      - POSTGRES_PASSWORD=MYPASSWORD
    ports:
      - "5442:5432"

  watchtower:
    container_name: watchtower
    restart: always
    image: v2tec/watchtower
    networks:
      - proxyv6
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
    command: --interval 240
```

This setup has always worked wonderfully with the original Timescaledb image. The databases I created remained intact even after an image update. With timescaledb-docker-ha, however, the databases are now gone after every image update :-(
Am I missing something here? Do I have to make an additional setting for timescaledb-docker-ha?

Unfortunately, I can't get any meaningful logs at the moment because I usually notice the image update too late. But I'll see if I can reproduce it by switching off the watchtower container and performing a manual update.

In any case, thanks in advance for your help!

wmkamerman:
Same here - I just switched from `timescale/timescaledb-postgis:latest-pg12` to `timescale/timescaledb-ha:pg12.11-ts2.7.2-latest`, only to find my relations gone.

The weird thing is that my DB volume has not changed in size, it just seems to not 'find' the data anymore. 

Any tips?

CM000n:
I have now also managed to get some logs after an update:
```log
{"log":"The files belonging to this database system will be owned by user \"postgres\".\n","stream":"stdout","time":"2022-07-25T09:49:13.210503707Z"}
{"log":"This user must also own the server process.\n","stream":"stdout","time":"2022-07-25T09:49:13.210569963Z"}
{"log":"\n","stream":"stdout","time":"2022-07-25T09:49:13.21058007Z"}
[...]

Issue #272, opened 2022-06-24
Image not fully compatible with standard postgres images


ryanobjc:
Hey folks, the primary documentation suggests that this image is a derivation from the official postgres images, but it does not appear to be true.  It's not technically true (although looks like you're attempting to replicate much of the same functionality), and it's not functionally true:  the timescaledb-ha image does not work with kubegres.io.

It would be nice if the image really was fully compatible with the postgres image one, and two kubegres.io in specific.  Looks like the main issue is you are changing the run user in Docker to 'postgres' rather than leaving it as 'root' which makes it difficult for secondary tooling to layer on if it requires root permissions to do things (such as create directories in mounted filesystems that didnt exist at image build time).

ryanobjc:
Additionally looks like this docker image doesn't declare a volume to mount AND puts the pg data in a completely different directory than the official postgres image, eg:
PGDATA=/home/postgres/pgdata/data

Whereas the official postgres image puts volumes here:
VOLUME /var/lib/postgresql/data

I was hoping to use this image to run timescale, but looks like I will have to build my own image from scratch instead to fix these issues, because kubegres is expecting things in the correct directories.

hiendaovinh:
@ryanobjc Do you have your image public somewhere?

nagaem:
I ran into this when trying to switch out the standard `timescaledb` image for this one in my docker-compose file. Trying to do so generates a permissions error:

```
chmod: changing permissions of '/var/lib/postgresql/data': Operation not permitted
The files belonging to this database system will be owned by user "postgres".
This user must also own the server process.
```

It might just be my inexperience, but it was non-obvious to me that the way to use this image was to change my volume mount path from the standard `/var/lib/postgresql/data` to `/home/postgres/pgdata/data` and drop my `PGDATA` env variable declaration from the docker compose file. If making the image compatible with standard postgres images is not an option, maybe this should be documented somewhere?

VMM-Mtech:
Before I saw this issue, I added a comment to #290. Both are related to the fact that different uid is used, and it cannot be given in env variable, because it has been hardcoded to several commands in Dockerfile.

danclimasevschi:
> I ran into this when trying to switch out the standard `timescaledb` image for this one in my docker-compose file. Trying to do so generates a permissions error:
> 
> ```
> chmod: changing permissions of '/var/lib/postgresql/data': Operation not permitted
> The files belonging to this database system will be owned by user "postgres".
[...]

PR #350, opened 2023-01-20
Allow the use of --user for docker containers

### LINKED ISSUE #345: Cannot use this image with a custom uid/gid and bind-mounts.
Docker allows telling a container as which user to run a process, using the `--user 1002:1002` argument or equivalent setting in docker-compose. This is especially important for docker hosts where the default uid/gid (1000:1000) belong to a user who should not have access to the data.

The official postgres docker image documentation suggests that passing this user flag and adding a read-only mapping of the hosts /etc/passwd will allow using a custom user. This does not work in timescale, as timescale has hardcoded the `postgres` user with the uid/gid 1000:1000.
https://github.com/timescale/timescaledb-docker-ha/blob/cc0deb86fcab07c25da7ac90af45ee0ff868ae62/Dockerfile#L39
https://github.com/timescale/timescaledb-docker-ha/blob/cc0deb86fcab07c25da7ac90af45ee0ff868ae62/Dockerfile#L519

Please provide guidance of bind-mounting the pgdata directory as another uid/gid than 1000:1000.

related issues:
- #272

graveland:
This needs to be combined with a volume mount if the user isn't uid/gid 1000/1000, as the user won't have permission to modify the default PGDATA location.

Use something like:
```
docker run \
    --user 1234:4567 \
    -e PGDATA=/home/pgdata \
    -v`pwd`/pgdata:/home/pgdata \
    $IMAGE
```

Fixed #345, and (partly?) #272. With this, #311 could be addressed potentially by specifying a path different than in that ticket, and might help a bit with #290 as well?


fsiler:
+1 for this patch.

Dockerfile
```diff
@@ -195,7 +195,7 @@ RUN cd /build && git clone https://github.com/timescale/timescaledb-docker && cd
 RUN cp -a /build/timescaledb-docker/docker-entrypoint-initdb.d /docker-entrypoint-initdb.d/
 # Add custom entrypoint to install timescaledb_toolkit
 COPY scripts/010_install_timescaledb_toolkit.sh /docker-entrypoint-initdb.d/
-RUN curl -s -o /usr/local/bin/docker-entrypoint.sh https://raw.githubusercontent.com/docker-library/postgres/${GITHUB_DOCKERLIB_POSTGRES_REF}/13/alpine/docker-entrypoint.sh
+COPY docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
```
feikesteenbergen, line 198: :+1:

docker-entrypoint.sh
```diff
@@ -0,0 +1,377 @@
+#!/usr/bin/env bash
```
feikesteenbergen, line 1: Shall we add a permalink at the top to clearly show where we git it from?

feikesteenbergen: `s/git/get/`

feikesteenbergen (approved):

docker-entrypoint.sh
graveland: Absolutely- good idea

fsiler:
Love that this got merged- can someone help me with when this will be used to publish official images/which versions it will be available in?

graveland:
Just today! Check out the pg15-latest, or pg14-latest (or the others from today as is appropriate: [timescaledb-ha](https://hub.docker.com/r/timescale/timescaledb-ha/tags?page=1))

PR #353, opened 2023-01-27
Initial pg15 support

graveland:
This changes the default PG_MAJOR to 15, and PG_VERSIONS to "15 14 13 12", as well as various
extension version upgrades.

`build_scripts` got updated, including fixing OSS_ONLY for timescaledb, and handling when
not all extension versions can be built for pg15.


Dockerfile
```diff
@@ -308,19 +308,19 @@ RUN if [ ! -z "${TIMESCALE_PROMSCALE_EXTENSIONS}" -a -z "${OSS_ONLY}" ]; then \
...
-        && mkdir /build/osm_extension \
-        && cd /build/osm_extension \ 
-        && for pg in ${PG_VERSIONS}; do \
-            if [ ${pg} -ge "14" ]; then \
+    set -ex; \
+    if [ -f "${REPO_SECRET_FILE}" -a -z "${OSS_ONLY}" -a -n "${TIMESCALE_OSM_EXTENSION}" ]; then \
+        mkdir /build/osm_extension; \
+        cd /build/osm_extension; \
+            for pg in ${PG_VERSIONS}; do \
+            if [ "${pg}" -ge "14" ]; then \
```
sebastianwebber, line 320: it's not the case to update for pg 15?

sebastianwebber (changes requested):
just a question regarding the pg version in the dockerfile.

Dockerfile
graveland: OSM supports pg 14+ as of 0.0.7, so this works.

manute (approved):

sebastianwebber (approved):
LGTM

PR #355, opened 2023-02-01
Refactor base HA image

graveland:
This is a large refactor that strips out all the things that aren't required for this image.

It also adds extra checks in the `cicd` scripts to make sure the requested extensions and
packages are actually present in the image. It does this by looking in the new `/.image_config`
file that contains the build parameters so we can do some introspection of the image.

The build scripts have been overhauled as well, including some matching for cargo-pgx
vs extension versions to optimize rust builds. Another major change is that requested
extensions first try to install `.deb` packages, and fall back to building them. The
timescaledb extension is always built though, as we want different build options than
what's currently in the debian packages.

Extension version and pg compatibility is described in `build_scripts/versions.yaml`
and that file should be the only thing that needs updating to add new versions of
extensions as they are released.

This also adds most of the required support for arm64 builds as well, but we need to
fix public arm64 build runners before fully building multiplatform images. We do currently
build images with `-amd64` tags, and create a manifest for the base tags, but only add
the one architecture to it for now.

There's also new github step summary information available after the build completes
that helps to easily see what's included in the image.

This currently pushes to `timescaledev/timescaledb-ha` instead of the main
`timescale/timescaledb-ha` docker hub repository. Changing that will be the last
thing to happen prior to merging this PR. This also won't be merged until other
downstream changes have been sorted out. Until then, we can build and test from
the `timescaledev/timescaledb-ha` images that this will build.

In order to speed things up, this branch's publish-images github action only builds pg15,
so adding 12-14 to the matrix would happen before merging as well.


build_scripts/shared_install.sh
```diff
@@ -0,0 +1,170 @@
...
+        if [ $err -eq 0 ]; then
+            log "installed $pkg-$version for pg$pg"
+        else
+            error "failed building $pkg-$version for pg$pg ($err)"
+        fi
+    done
+    PATH="$ORIGINAL_PATH"
+}
+
+install_promscale() {
```
manute, line 124: promscale is a frozen project now, not sure if it's worthy to include here

manute (commented):
There is a lot of bash code that has not got tests, sure it will work but not sure if we can make tests or is it too complex?

graveland:
> > and versions later than that are installed from pre-built debian packages.
> 
> Tricky with that is that we want debug symbols included in the libraries - that is the reason we compile them ourselves currently. `-DCMAKE_BUILD_TYPE=RelWithDebInfo`
> 
> The second thing we wanted to ensure is that the downgrade scripts are included (`DGENERATE_DOWNGRADE_SCRIPT=ON`), it looks like that is covered nowadays in the deb packages themselves.
> 
> The third thing we wanted to insure is the `INSTALL_METHOD` is distinct from timescaledb installed from packages.
> 
> ```
>         ./bootstrap \
>             -DTAP_CHECKS=OFF \
>             -DWARNINGS_AS_ERRORS=off \
>             -DCMAKE_BUILD_TYPE=RelWithDebInfo \
>             -DREGRESS_CHECKS=OFF \
>             -DGENERATE_DOWNGRADE_SCRIPT=ON \
>             -DPROJECT_INSTALL_METHOD="${INSTALL_METHOD}${OSS_ONLY}"
> ```

Sounds good. It's an easy tweak to build always. Looking at the debian packages, there are only a few with corresponding `-dbgsym` packages. I'll change that.

graveland:
> There is a lot of bash code that has not got tests, sure it will work but not sure if we can make tests or is it too complex?

I can't say I ever really made tests for shell scripts? Between running the `cicd/install_checks` during build time, and CI time, and the github step summaries showing what actually got installed, I figured it was good enough? It's certainly more complex than before, but we also were missing things without manually looking inside the resulting images.

If people want tests, or if there is any more complexity at all, I'd switch languages.

graveland:
The current procedure for adding a new version of an extension includes modifying `build_scripts/shared_versions.sh` to update the appropriate `pgx_X_versions` structure if you're changing one of those, and/or modifying the `supported_X` functions that determine if a package version + pg version should result in an installation/build.

Streamlining that would be a win. I'll think a bit on it more.

graveland:
> The current procedure for adding a new version of an extension includes modifying `build_scripts/shared_versions.sh` to update the appropriate `pgx_X_versions` structure if you're changing one of those, and/or modifying the `supported_X` functions that determine if a package version + pg version should result in an installation/build.
> 
> Streamlining that would be a win. I'll think a bit on it more.

The latest push includes `build_scripts/versions.yaml` and is now the only place that you need to edit when adding a new version of an extension. This also adds the ability to use `all` and `latest` when specifying versions of extensions that you want to build, and the `Makefile` defaults to `all`. To add timescaledb 2.9.3 for example, add one line to the file, and it will be included. For faster builds / testing, the make target `latest` uses to `latest` version specification for `timescaledb`, `timescaledb_toolkit`, and `promscale`.

Makefile
```diff
@@ -166,60 +228,106 @@ build-oss: build
...
+build-oss: # build an OSS-only image
+build-oss: OSS_ONLY=true
+build-oss: DOCKER_TAG_POSTFIX=-oss
+build-oss: release
+
+.PHONY: publish-combined-builder-manifest
+publish-combined-builder-manifest: # publish a combined builder image manifest
+	@echo "Creating manifest $(DOCKER_BUILDER_URL) that includes $(DOCKER_BUILDER_URL)-amd64"
+	amddigest_image="$$(./fetch_tag_digest $(DOCKER_BUILDER_URL)-amd64)"
+	echo "AMD: $$amddigest_image"
```
sebastianwebber, line 242: Arm info is missing. is this the case to add it?

sebastianwebber (approved):
two small comments. LGTM

.github/workflows/publish_images.yaml
```diff
@@ -0,0 +1,120 @@
+name: Publish images
+
+on:
+  workflow_dispatch:
+
+  push:
+    branches:
+      - main
+      - brent/base
```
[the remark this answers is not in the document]
graveland: yup

Makefile
graveland: It is. Arm was working 100%, but we need to sort out a secure arm64 builder infrastructure. I've got a PoC for that, but I need to sort out the webhooks for it from GH before re-introducing arm64.

alexeyklyukin:
Tricky to review, but we can try it on dev. A declarative way to describe build options and extensions/versions is a huge win. I'd add a step to list build docker images, which helps find the tag to put into the deployer config.

.github/workflows/publish_images.yaml
```diff
@@ -0,0 +1,120 @@
...
+    paths-ignore:
+      - "*.md"
+
+  schedule:
+    - cron: '0 7 * * 2'
```
feikesteenbergen, line 14: For confirmation: we'll be publishing an image every Tuesday at 07:00 UTC, that's what we want?

README.md
```diff
@@ -10,6 +10,8 @@ This directory contains everything that allows us to create a Docker image with
 
 Currently, our base image is Ubuntu, as we require glibc 2.33+.
 
+It's currently pushing the resuting images to: https://hub.docker.com/r/timescaledev/timescaledb-ha
```
feikesteenbergen, line 13:
```suggestion
It's currently pushing the resulting images to: https://hub.docker.com/r/timescale/timescaledb-ha
```

feikesteenbergen: I think we push to `/timescale/` since commit 0c54000b3eb213fffc96d37ba98d9758a6adbddf

build_scripts/shared.sh
```diff
@@ -0,0 +1,192 @@
+#!/bin/bash
```
feikesteenbergen, line 1: I like this!

build_scripts/versions.yaml
```diff
@@ -0,0 +1,116 @@
+# We need to describe various things for each version of each extension:
```
feikesteenbergen, line 1: :100:

cicd/version_info.sql
```diff
@@ -9,7 +9,7 @@ WITH versions(name, version) AS (
     FROM
         pg_available_extensions
     WHERE
-        name IN ('timescaledb', 'postgis', 'pg_prometheus', 'hotforge', 'timescale_prometheus_extra', 'promscale', 'timescaledb_toolkit', 'timescale_analytics', 'timescaledb_cloudutils', 'timescaledb_osm')
```
feikesteenbergen, line 12: Why are we removing these extensions from the where clause?
(If the image doesn't contain the extension, the resulting rows would not include these anyway).

feikesteenbergen (approved):
LGTM, a few nits, nothing major

Dockerfile
```diff
@@ -31,47 +24,73 @@ ARG PG_MAJOR=13
...
 
 # The following tools are required for some of the processes we (TimescaleDB) regularly
 # run inside the containers that use this Docker Image
 # awscli is useful in many situations, for example, to list backup buckets etc
-RUN apt-get install -y less jq strace procps awscli
+RUN set -eux; \
+    apt-get update; \
+    apt-get upgrade -y; \
+    apt-get install -y \
+        less jq strace procps awscli vim-tiny gdb gdbserver dumb-init daemontools \
```
alexeyklyukin, line 81: 💯 for vim !

alexeyklyukin (approved):
LGTM

.github/workflows/publish_images.yaml
graveland: The time picked was arbitrary, and it stepped on me during testing a couple times so I bounced it around :)

The idea is to automatically pick up updated .debs to automatically include security fixes. I'd like it scheduled, but I don't care at all about when.

README.md
graveland: yup, is pushing to timescaledev just so that it can run concurrently with the master branch, and would be a thing that changes just before merging.

cicd/version_info.sql
graveland: Just to remove internal timescale parts from this public image.

graveland:
> Tricky to review, but we can try it on dev. A declarative way to describe build options and extensions/versions is a huge win. I'd add a step to list build docker images, which helps find the tag to put into the deployer config.

We wouldn't be using the results of this in the deployer, and the full image URLs/tags are in the action step summary for easy access :)

build_scripts/shared.sh
sebastianwebber: 💯

.github/workflows/publish_images.yaml
feikesteenbergen: 👍

cryptobench:
Looking forward to this as i'm interested in the ARM image for local development so I can achieve env parity. Thank you for the work you're putting in!

Dockerfile
```diff
@@ -31,47 +24,73 @@ ARG PG_MAJOR=13
...
-RUN apt-get install -y gdb gdbserver
+# forbid creation of a main cluster when package is installed
+RUN sed -ri 's/#(create_main_cluster) .*$/\1 = false/' /etc/postgresql-common/createcluster.conf
 
 # The next 2 instructions (ENV + RUN) are directly copied from https://github.com/rust-lang/docker-rust/blob/d534735bae832da4c60ddf799a8dfbefa9939020/1.67.0/bullseye/Dockerfile
 ENV RUSTUP_HOME=/usr/local/rustup \
     CARGO_HOME=/usr/local/cargo \
     PATH=/usr/local/cargo/bin:$PATH \
-    RUST_VERSION=1.65.0
+    RUST_VERSION=1.66.1
```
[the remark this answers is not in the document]
graveland: Absolutely, and it does that mostly. Timescaledb, promscale, and toolkit builds happen in `build_scripts/install_extensions`. If it can find the artifact for rust things, it installs them, otherwise it'll build them. Timescaledb is always built though, since we use different compilation arguments than the .debs that are available.

build_scripts/shared.sh
thedodd: Solid script!

build_scripts/shared_install.sh
```diff
@@ -0,0 +1,243 @@
...
+    FROM
+        sql_time_bucket_fn AS s
+    WHERE
+        s.oid = pg_proc.oid;
+END;
+$dynsql$;
+__SQL__
+        done # for file
+    done     # for pg
+}
```
thedodd, line 243: @graveland I think your bash rivals that of Cam (our ol' Aussie chap).

cicd/shared.sh
```diff
@@ -0,0 +1,306 @@
...
+                error "pg$pg extension postgis-$ver not found: $status"
+            fi
+        done
+    fi
+
+    for extname in $PG_WANTED_EXTENSIONS; do
+        record_ext_version "$extname" "$pg" ""
+        IFS=\| read -rs version status <<< "$(dpkg-query -W -f '${version}|${status}' "postgresql-$pg-$extname")"
+        if [ "$status" = "install ok installed" ]; then
+            #log "found pg$pg extension $extname-$version"
```
thedodd, line 250: Looks like there is a fair number of log statements which have been commented out in this script. Not sure if this is just temporary to reduce noise while debugging, or if we actually want to remove them.

Thoughts? Should we enable the commented log lines?

thedodd (approved):
Great stuff!

cicd/shared.sh
graveland: `log` sends stuff to the github step summary, so we don't want to flood that. I enable for testing, but left them commented out as I've bounced between using them and not, depending on what I was troubleshooting.

I'm fine with zapping them though, as we really don't want them outputting all that extra stuff. I don't think it's terribly useful to have those statements on any output on a regular basis.

graveland: Gone

aries1980:
Thanks, mate!

leppaott:
`the -latest suffix here indicates the latest docker build, not the latest commit. In particular, images built from a tag will be published with the -latest suffix in addition to the tag-based suffix.
`

From Main README are the image tagging prefixes still valid? How about this '-all'?

graveland:
`-all` means that the specified image includes all the pg versions (pg15-all includes pg12-15). Images with the `-latest` tag aren't actually getting published as of this PR merge. I'll have to update that README in a PR ASAP!

Now that the builds are scheduled, there's a slight change. Taking `pg15.2-ts2.10.3` as an example, between two builds, pg and timescale won't change, but it's possible that the postgis version does change since we're getting upstream fixes from debian and pgdg automatically. If somebody doesn't want the image to change at all, then I'd suggest using the SHA-based tags `sha256:8a32cbd9709f5cb9bc56e39a7832e6285ecbfc2f978e58165c946c87c36a8d2e`.

If you want to suggest any changes to how the tags are pushed, let us know!

polarn:
@graveland I think the introduction of `--platform` for the docker build introduced a small problem for people running in for example Rancher Desktop on Silicon M1/M2 Macs. Before this, the image produced was of the old non multi-arch type and now it is of multi arch (with only one arch, amd64). It confuses Rancher to try to pull the arm64 image, but there is no such image (yet). The behaviour before with the "single" architecture image was to pull that.

Not thinking you should do anything about it, just wanted to mention it if others see this in their Kubernetes:

```
Failed to pull image "timescale/timescaledb-ha:pg15.2-ts2.10.3": rpc error: code = Unknown desc = no matching manifest for linux/arm64/v8 in the manifest list entries
```

A work-around for users could maybe be to point out the image digest instead of tag, or just pulling the image manually to the dockerd.

If you are curious to see the difference in image types, check the output of these:

```
docker manifest inspect --verbose timescale/timescaledb-ha:pg15.2-ts2.10.3 
docker manifest inspect --verbose timescale/timescaledb-ha:pg15.2-ts2.10.1-latest
```

You will see an array with one element for the first command and you will see the element directly on the root for the second image.

PR #391, opened 2023-05-19
Add pgxn and pg-trunk to make adding pg extensions easier

graveland:


sebastianwebber (approved):
this is awesome! 

LGTM

Dockerfile
```diff
@@ -153,6 +152,21 @@ RUN set -eux; \
...
+
+RUN set -eux; \
+    for pg in ${PG_VERSIONS}; do \
+        for pkg in h3 orafce pgTAP; do \
+            pgxn install --pg_config "/usr/lib/postgresql/${pg}/bin/pg_config" "$pkg"; \
+        done; \
+        for pkg in pg_uuidv7 intagg; do \
+            trunk install --pg-config "/usr/lib/postgresql/${pg}/bin/pg_config" "$pkg"; \
+        done; \
+    done
```
feikesteenbergen: `intagg` is a [contrib module](https://www.postgresql.org/docs/current/intagg.html) and already available:
```
tsdb=> create extension intagg with schema public;
CREATE EXTENSION
tsdb=> \dx+ intagg
      Objects in extension "intagg"
            Object description            
------------------------------------------
 function int_agg_final_array(internal)
 function int_agg_state(internal,integer)
 function int_array_aggregate(integer)
 function int_array_enum(integer[])
(4 rows)
```

```suggestion
        for pkg in pg_uuidv7; do \
            trunk install --pg-config "/usr/lib/postgresql/${pg}/bin/pg_config" "$pkg"; \
        done; \
    done
```

Dockerfile
```diff
@@ -153,6 +152,21 @@ RUN set -eux; \
...
+# Add a couple 3rd party extension managers to make extension additions easier
+RUN set -eux; \
+    apt-get install -y pgxnclient; \
+    cargo install pg-trunk
+
+RUN set -eux; \
+    for pg in ${PG_VERSIONS}; do \
+        for pkg in h3 orafce pgTAP; do \
+            pgxn install --pg_config "/usr/lib/postgresql/${pg}/bin/pg_config" "$pkg"; \
+        done; \
```
feikesteenbergen: Both `orafce` and `pgtap` exist in the pgdg debian repos, so we don't need to install them here like this

```suggestion
        for pkg in h3; do \
            pgxn install --pg_config "/usr/lib/postgresql/${pg}/bin/pg_config" "$pkg"; \
        done; \
```

feikesteenbergen (approved):

PR #405, opened 2023-07-17
Check for more extensions

graveland:
h3 wasn't actually being built properly for all pg versions, so we have to build it ourselves and pg_uuidv7 had a release that wasn't building for pg-trunk so installing via pgxn works better.


sebastianwebber (approved):
LGTM

PR #411, opened 2023-08-28
Simplify pgbouncer_exporter installation

graveland:
Use the same style as the rest of the Dockerfile and simplify the use of dpkg --print-architecture, removing the remaining embedded amd64 statements.


alexeyklyukin (approved):
Nice!

PR #413, opened 2023-08-31
Install pgvector packages

alexeyklyukin:
Add pgvector extension using Ubuntu packages


graveland:
pgvector is already installed via building, so if it's finally available as a package, that should be deleted too

feikesteenbergen (approved):

PR #423, opened 2023-10-12
Update to TimescaleDB 2.12.1

alexeyklyukin:
Update to TimescaleDB 2.12.1

Remove custom Patroni patch, as it should be part of the
official release.


Dockerfile
```diff
@@ -200,7 +200,6 @@ COPY --chown=postgres:postgres build_scripts /build/scripts/
 # This need to be done after the PostgreSQL packages have been installed,
 # to ensure we have the preferred libpq installations etc.
 RUN apt-get install -y python3-etcd python3-requests python3-pystache python3-kubernetes python3-pysyncobj patroni
-RUN cd /usr/lib/python3/dist-packages/patroni && git apply /build/scripts/patroni_issue_2837.patch
```
feikesteenbergen, line 203: Great! We can also `rm` the file?

feikesteenbergen (approved):

Dockerfile
alexeyklyukin: We can and I did remove it in commit https://github.com/timescale/timescaledb-docker-ha/pull/423/commits/a63c847d056ef7b27d6368af543fa32c8a3d94ee

PR #431, opened 2023-12-15
Including pgvecto.rs plugin

dereulenspiegel:
This change adds the possibility to add [pgvecto.rs](https://github.com/tensorchord/pgvecto.rs) as an additional extension. I am aware that pgvector is already included, but some projects like [immich](https://github.com/immich-app/immich) have started using pgvecto.rs instead. I would like to continue this image as my "go-to" postgres image.

Open Tasks

* [x] Added vectors extension to `shared_preload_libraries` so that the extension can be used immediately


dereulenspiegel:
The way I added this extension to `shared_preload_libraries` is maybe not optimal, so I am to suggestions how to improve on this.

Dockerfile
```diff
@@ -258,7 +274,7 @@ RUN set -ex; \
 
 RUN for file in $(find /usr/share/postgresql -name 'postgresql.conf.sample'); do \
         # We want timescaledb to be loaded in this image by every created cluster
-        sed -r -i "s/[#]*\s*(shared_preload_libraries)\s*=\s*'(.*)'/\1 = 'timescaledb,\2'/;s/,'/'/" $file \
+        sed -r -i "s/[#]*\s*(shared_preload_libraries)\s*=\s*'(.*)'/\1 = 'vectors,timescaledb,\2'/;s/,'/'/" $file \
```
svenklemm, line 277: Won't this result in postgres not starting when vectors is missing?

dereulenspiegel: Yes, you are right, totally missed that. Is there a recommended way to configure `shared_preload_libraries` for the additional extensions?

graveland: I think that should generally be left to the users of the image to supply a postgresql.conf that includes whatever they want.

Makefile
```diff
@@ -16,6 +16,7 @@ PG_AUTH_MON?=v2.0
 PG_STAT_MONITOR?=2.0.3
 PG_LOGERRORS?=18d9795
 PGVECTOR?=v0.5.1
+PGVECTO_RS?=0.1.11
```
graveland, line 19: Since 0.1.13 was just released, I'd suggest updating to that one and testing. Might as well start with the latest!

dereulenspiegel: Absolutely. This was released after I created the PR, but the PR is now updated to version `0.1.13`. Thanks for the hint to the latest release.

Dockerfile
dereulenspiegel: Good point. Is there documentation somewhere, that I might have missed, which would need updating in this regard?

graveland: Not really? This image follows the same basic methods as the [postgresql docker image database configuration](https://www.docker.com/blog/how-to-use-the-postgres-docker-official-image/#4-Database-configuration) where we pass the command-line arguments to `postgres`, so you can add `-c` to set run-time parameters, or you could use docker's `-v` to map a configuration file to the container. I didn't really find where overriding `shared_preload_libraries` is documented.

I did test that this works though:
`docker run --rm -it -e POSTGRES_HOST_AUTH_METHOD=trust timescale/timescaledb-ha:pg16-ts2.13 -c shared_preload_libraries=pg_stat_statements,timescaledb`

graveland (approved):
lgtm!

Since CI never ran this, we can fix problems with a follow-on PR if needed.

Dockerfile
dereulenspiegel: Just wanted to check if I missed updating relevant documentation. Thanks :)

graveland:
Do you want me to merge this and validate/fix build problems?

dereulenspiegel:
Merging this would be awesome. I already ran a first test, that I can create this extension in a PG15 database, but for thorough testing I probably won't have the required time at least until after Chirstmas. But merging this shouldn't break anything.

Dockerfile
```diff
@@ -206,6 +206,22 @@ RUN set -ex; \
...
+ARG PGVECTO_RS
+RUN set -ex; \
+    if [ -n "${PGVECTO_RS}" ]; then \
+        for pg in ${PG_VERSIONS}; do \
+            # Vecto.rs only support PostgreSQL 14 and upwards
+            if [ $pg -gt 13 ]; then \
+                curl --silent \
+                    --location \
+                    --output /tmp/vectors.deb \
+                    "https://github.com/tensorchord/pgvecto.rs/releases/download/v${PGVECTO_RS}/vectors-pg${pg}-v${PGVECTO_RS}-$(uname -m)-unknown-linux-gnu.deb" && \
```
graveland, line 218: I missed this, they changed the artifact naming:

```suggestion
                    "https://github.com/tensorchord/pgvecto.rs/releases/download/v${PGVECTO_RS}/vectors-pg${pg}_v${PGVECTO_RS}_$(dpkg --print-architecture).deb" && \
```

dereulenspiegel: Thanks for the hint, missed this as well.

graveland:
It's in the images now :) https://github.com/timescale/timescaledb-docker-ha/pull/432

dereulenspiegel:
Awesome, thanks :)

PR #450, opened 2024-03-11
Use an older tzdata file that's still available

### LINKED ISSUE #448: Issue with tzdata dependency (again)
I am trying to build using this command:

```
PLATFORM=aarch64 \
DOCKER_REGISTRY=myregistry.example.com \
DOCKER_REPOSITORY=infrastructure/docker-images/timescaledb-ha \
DOCKER_TAG_POSTFIX=-aarch64 \
make build
```

I am getting this error:

```
#27 [builder 23/62] RUN apt-get install -y --allow-downgrades tzdata="2023c-*"
#27 0.244 Reading package lists...
#27 1.212 Building dependency tree...
#27 1.460 Reading state information...
#27 1.468 Package tzdata is not available, but is referred to by another package.
#27 1.468 This may mean that the package is missing, has been obsoleted, or
#27 1.468 is only available from another source
#27 1.468
#27 1.472 E: Version '2023c-*' for 'tzdata' was not found
#27 ERROR: process "/bin/sh -c apt-get install -y --allow-downgrades tzdata=\"2023c-*\"" did not complete successfully: exit code: 100
------
 > [builder 23/62] RUN apt-get install -y --allow-downgrades tzdata="2023c-*":


1.468 Package tzdata is not available, but is referred to by another package.
1.468 This may mean that the package is missing, has been obsoleted, or
1.468 is only available from another source
1.468
1.472 E: Version '2023c-*' for 'tzdata' was not found
------
Dockerfile:156
--------------------
 154 |
 155 |     ## TODO: There's currently a build-dependency problem related to tzdata, remove this when it's resolved
 156 | >>> RUN apt-get install -y --allow-downgrades tzdata="2023c-*"
 157 |
 158 |     # We install the PostgreSQL build dependencies and mark the installed packages as auto-installed,
--------------------
ERROR: failed to solve: process "/bin/sh -c apt-get install -y --allow-downgrades tzdata=\"2023c-*\"" did not complete successfully: exit code: 100
make: *** [Makefile:253: build] Error 1
```

I see that this `apt install` was added recently in #437, but it looks like tzdata version 2023c might no longer be listed in the package indexes for Ubuntu 22.04.

Any thoughts on how to fix this?

graveland:
This also adds a bit more context to the problem.

https://salsa.debian.org/postgresql/postgresql/-/commit/b995beb3cd1c2b8834605007227b3cedab6462e4 says that this is required only for testing, so I think we're safe to use the current work-around. The previous tzdata file isn't available anymore, but one older file is still around. If the old tzdata file disappears too, we'll have to do something more drastic.

This should fix #448

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -152,8 +152,11 @@ ENV BUILD_PACKAGES="binutils cmake devscripts equivs gcc git gpg gpg-agent libc-
 RUN apt-get install -y ${BUILD_PACKAGES}
 RUN apt-mark auto ${BUILD_PACKAGES}
 
-# TODO: There's currently a build-dependency problem related to tzdata, remove this when it's resolved
-RUN apt-get install -y --allow-downgrades tzdata="2023c-*"
+# https://salsa.debian.org/postgresql/postgresql/-/commit/b995beb3cd1c2b8834605007227b3cedab6462e4
+# This looks like a build-/test-only dependency, and they expect us to use the real tzdata when actually running.
+# TODO: keep watching this to see if they remove the limitation. If the old tzdata becomes unavailable, we'll have to
+# do something more drastic.
+RUN apt-get install -y --allow-downgrades tzdata="2022a-*"
 
 # We install the PostgreSQL build dependencies and mark the installed packages as auto-installed,
 RUN set -eux; \


adolsalamanca (approved):

alexeyklyukin (approved):
Looks good!

PR #442, opened 2024-02-08
Added Barman Cloud for CloudNativePG compatibility

itay-grudev:
Closes #422

PR #459, opened 2024-04-30
Bug Fix: Missing Barman Cloud entrypoints

itay-grudev:
In #442 I added the `python3-barman` package in anticipation for the `ImageCatalog` feature in CloudNativePG that would allow running the official Timescale image in CNPG.

That PR though has two issues:
1. Unlike `python-pip`, `apt-get` does not install package binaries, required for interaction with `barman-cloud`.
2. It did not add necessary dependencies for Azure and Google cloud support.

I tried implementing it without `python-pip`, but the Azure Storage package `python3-azure-storage` is over `600MB` which completely defeats the purpose of the size optimization. With `pip` the `barman` package only fetches `azure-blob-storage` and this results in around `300MB` smaller image, compared to the deb package only approach.

I suspect that once I add `python3-pip` it may be beneficial for other as well to switch to `pip` packages instead of Debian packages as:
* More targeted packages have a smaller size impact and less bloat
* Newer versions of packages are available via `pip`.

In fact if you request it, I will move the `apt-get install -y python3-pip` to the beginning of the Dockerfile so it is available to use earlier. I only kept it here so the Docker image is easier to build and use more cache.

P.S. Special thanks to @NiccoloFei for his assistance in debugging this.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -230,7 +230,8 @@ RUN apt-get install -y python3-etcd python3-requests python3-pystache python3-ku
 
 # Barman cloud
 # Required for CloudNativePG compatibility
-RUN apt-get install -y python3-barman
+RUN apt-get install -y python3-pip && \
+    pip3 install --no-cache-dir 'barman[cloud,azure,snappy,google]'
 
 RUN apt-get install -y timescaledb-tools
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -146,6 +146,9 @@ RUN set -eux; \
     echo 'en_US.UTF-8 UTF-8' > /usr/share/i18n/SUPPORTED; \
     localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8
 
+# We install pip3, as we need it for some of the extensions. This will install a lot of dependencies, all marked as auto to help with cleanup later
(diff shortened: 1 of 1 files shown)


itay-grudev:
@graveland Can I trouble you with a review, again?

itay-grudev:
@feikesteenbergen @graveland I would like to release support for TimescaleDB in the official CloudNativePG chart next week. We have everything else prepared, but this PR.

Would you be able to take allocate some time for a review?

graveland (approved):
I like the idea of moving the installation of pip itself to the beginning as there's another PR coming tbat uses it too.

To bad this adds so much storage!

itay-grudev:
@graveland Would you like me to make that change now?

graveland:
> @graveland Would you like me to make that change now?

If you don't mind, thanks!

itay-grudev:
I scanned through the Dockerfile and [this](https://github.com/timescale/timescaledb-docker-ha/pull/459/commits/1cc69be165a21440303db926f2a714b1c679f62b) seems as a good spot. I also explicitly added `python3` so it's not marked as auto, as it is required at runtime.

itay-grudev:
@graveland 

P.S. I can't marge this, but ping me if you would like me to make any other changes.

graveland:
All good, thanks!

PR #462, opened 2024-05-23
Include pgai extension

jgpruitt:


Dockerfile
```diff
@@ -190,6 +190,22 @@ RUN set -ex; \
         done; \
     fi
 
+# pgai is an extension for artificial intelligence
+ARG PGAI
```
graveland, line 194: All the other once have `_VERSION`, so for consistency, please:
```suggestion
ARG PGAI_VERSION
```

graveland (approved):
I added a couple trivial comments, but otherwise it looks good to me :)

One other comment: you should add a check in `cicd/shared.sh`, probably in the `check_others` function to validate that pgai was actually installed when it was requested.

Dockerfile
jgpruitt: fixed

Dockerfile
```diff
@@ -190,6 +190,22 @@ RUN set -ex; \
...
+RUN set -ex; \
+    if [ -n "${PGAI}" ]; then \
+        git clone https://github.com/timescale/pgai.git /build/pgai; \
+        cd /build/pgai; \
+        git checkout "${PGAI}"; \
+        git reset HEAD --hard; \
+        apt-get install -y python3-full python3-pip; \
```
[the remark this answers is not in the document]
jgpruitt: Ok. I separated the python installs and the git/build commands. I checked, and it looks like I won't need python3-full, so I removed it. Also, I'm using a requirements.txt file now, so the python installs need to stay in this section.

Dockerfile
```diff
@@ -190,6 +190,22 @@ RUN set -ex; \
...
+    if [ -n "${PGAI_VERSION}" ]; then \
+        git clone https://github.com/timescale/pgai.git /build/pgai; \
+        cd /build/pgai; \
+        git checkout "${PGAI_VERSION}"; \
+        git reset HEAD --hard; \
+        for pg in ${PG_VERSIONS}; do \
+            cp /build/pgai/ai--*.sql "$(/usr/lib/postgresql/${pg}/bin/pg_config --sharedir)/extension/"; \
+            cp /build/pgai/ai.control "$(/usr/lib/postgresql/${pg}/bin/pg_config --sharedir)/extension/"; \
+        done; \
+        apt-get install -y python3-pip; \
```
graveland, line 205: If you rebase on master, `python3-pip` is already installed on line 150, so you could drop that.

Dockerfile
```diff
@@ -457,6 +473,9 @@ RUN /build/scripts/install_extensions versions > /.image_config; \
     echo "BUILDER_URL=\"${BUILDER_URL}\"" >> /.image_config; \
     echo "BUILD_DATE=\"$(date -Iseconds)\"" >> /.image_config
 
+#    echo "PGAI_VERSION=\"${PGAI_VERSION}\"" >> /.image_config; \
```
graveland, line 476: Should uncomment and move this up a few lines, just after PGBACKREST or so?

graveland (approved):
requirements.txt is a great improvement!

Dockerfile
jgpruitt: ah yes. thanks

PR #466, opened 2024-06-04
Only support pgai on pg16+

jgpruitt:
We only want to support the pgai extension on postgres version 16 or greater.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -196,14 +196,16 @@ RUN set -ex; \
 # pgai is an extension for artificial intelligence workloads
 ARG PGAI_VERSION
 RUN set -ex; \
-    if [ -n "${PGAI_VERSION}" ]; then \
+    if [ "${PG_MAJOR}" -gt 15 ] && [ -n "${PGAI_VERSION}" ]; then \
         git clone https://github.com/timescale/pgai.git /build/pgai; \
         cd /build/pgai; \
         git checkout "${PGAI_VERSION}"; \
         git reset HEAD --hard; \
         for pg in ${PG_VERSIONS}; do \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -468,7 +470,7 @@ RUN /build/scripts/install_extensions versions > /.image_config; \
     echo "PG_AUTH_MON=\"${PG_AUTH_MON}\"" >> /.image_config; \
     echo "PGBOUNCER_EXPORTER_VERSION=\"${PGBOUNCER_EXPORTER_VERSION}\"" >> /.image_config; \
     echo "PGBACKREST_EXPORTER_VERSION=\"${PGBACKREST_EXPORTER_VERSION}\"" >> /.image_config; \
-    echo "PGAI_VERSION=\"${PGAI_VERSION}\"" >> /.image_config; \
+    if [ "${PG_MAJOR}" -gt 15 ]; then echo "PGAI_VERSION=\"${PGAI_VERSION}\"" >> /.image_config; fi; \
```
graveland, line 473: I'd suggest not guarding this- if PGAI_VERSION isn't set, we'll get an empty version and that's good too.

cicd/shared.sh
```diff
@@ -228,7 +228,7 @@ check_others() {
     fi
 
     record_ext_version ai "$pg" ""
-    if [ -n "$PGAI_VERSION" ]; then
+    if [ -n "$PGAI_VERSION" ] && [ "$pg" -gt 15 ]; then
```
graveland, line 231: I'm partial to this since we're using bash, but feel free to ignore me :)

```suggestion
    if [[ -n "$PGAI_VERSION" && "$pg" -gt 15 ]]; then
```

graveland (approved):
lgtm

Dockerfile
jgpruitt: done

PR #472, opened 2024-06-21
Add arm for pgvectorscale

cevian:


build_scripts/shared_install.sh
```diff
@@ -245,7 +245,10 @@ __SQL__
 }
 
 install_pgvectorscale() {
-    local version="$1" pg pkg=pgvectorscale unsupported_reason
+    local version="$1" pg pkg=pgvectorscale unsupported_reason arch_deb="$ARCH"
+    if [ "$arch_deb" = aarch64 ]; then
```
graveland, line 249: The top-level `shared.sh` includes both `shared_install.sh` and `shared_versions.sh`, which is where ARCH is set. This seems like possibly a useful thing to have? Possibly moving this to `shared_versions.sh` so other things could use it in the future? At the time this was written, aarch64 was more useful, but there's always been the problem of different things calling it different names. The problem is that I don't really have a good name, and ARCH_DEB doesn't seem quite right. ARCH_ALT? CANONICAL_ARCH?

graveland (approved):
Thanks! Added a nit. Feel free to ignore it if you want :)

build_scripts/shared_install.sh
cevian: Yeah my inclination is to keep one canonical name at the top "$ARCH" which uses our pre-defined names, and have subsystems do their own translation. If we find this pattern recurring we can make a function that does the translation

cevian: another words I wouldnt define multiple global names.

graveland: Totally reasonable and fine by me. The only reason I brought it up is that it seems to _always_ be a problem and I haven't come up with a good solution. x86_64 vs amd64 and aarch64 vs arm64 are just a pain in the butt!

PR #481, opened 2024-07-22
Add postgres_exporter to docker image

saolof:
The missing exporter in the trio. Not mandatory, but still quite useful to have when dealing with a lagging replica for example.

Ultimately whether it should be part of the image or not is an engineering tradeoff, feel free to reject this for any reason. The pgbackrest exporter looks directly at the pgbackrest volume and benefits from being in the same image. The other two could be in a separate image but are convenient to include in order to use the same image for everything.


graveland (approved):
Thanks!

I'm on the fence about including it, but maybe the decision to not make this image even larger gets to be pushed off a bit more :)

PR #485, opened 2024-08-19
Add xfsprogs to the image

minkimipt:


minkimipt:
xfs tools are not needed inside the image. We'll need them on the host do do something useful with them

PR #488, opened 2024-09-04
pgai version 0.4.0

jgpruitt:
Adds pgai extension version 0.4.0

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ PG_MAJOR?=16
 PG_VERSIONS?=
 
 # Additional PostgreSQL extensions we want to include with specific version/commit tags
-PGAI_VERSION?=v0.3.0
+PGAI_VERSION?=extension-0.4.0
 PGVECTORSCALE_VERSIONS?=all
 POSTGIS_VERSIONS?=3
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -11,7 +11,7 @@ PG_MAJOR?=16
 PG_VERSIONS?=
 
 # Additional PostgreSQL extensions we want to include with specific version/commit tags
-PGAI_VERSION?=v0.3.0
+PGAI_VERSION?=extension-0.4.0
```
graveland, line 14: Why isn't there a v0.4.0 tag? That would remove the need for a lot of this PR, and would keep it more in-line with how things are usually done?

graveland (approved):
I'd suggest a v0.4.0 tag, then most of this PR can be reverted and it'll be in-line with how things are normally done. Approving to unblock if you really want it this way.

Makefile
jgpruitt: the pgai repo is now a mono-repo with both an extension and a python library. the python library is being tagged independently as `pgai-x.x.x` and the extension as `extension-x.x.x`

PR #489, opened 2024-09-12
Add builder tags to allow minor pg version usage

graveland:


Makefile
```diff
@@ -266,11 +266,14 @@ publish-combined-builder-manifest: # publish a combined builder image manifest
 	amddigest_image="$$(./fetch_tag_digest $(DOCKER_BUILDER_URL)-amd64)"
 	armdigest_image="$$(./fetch_tag_digest $(DOCKER_BUILDER_URL)-arm64)"
 	echo "AMD: $$amddigest_image ARM: $$armdigest_image"
-	docker manifest rm "$(DOCKER_BUILDER_URL)" >& /dev/null || true
-	docker manifest create "$(DOCKER_BUILDER_URL)" --amend "$$amddigest_image" --amend "$$armdigest_image"
-	docker manifest push "$(DOCKER_BUILDER_URL)"
-	echo "pushed $(DOCKER_BUILDER_URL)"
-	echo "Pushed $(DOCKER_BUILDER_URL) (amd:$$amddigest_image, arm:$$armdigest_image)" >> "$(GITHUB_STEP_SUMMARY)"
+	for tag in pg$(PG_MAJOR) pg$(VAR_PGMAJOR); do
```
minkimipt, line 269: I'm not sure if the name `VAR_PGMAJOR` is accurate here because it seems that it will contain the version in format 16.4, which is the full version. Is there a better name for them when they are used together?

minkimipt (approved):

Makefile
alexeyklyukin, line 269: so we post manifests tags to keep compatibility with the older image (i.e. pg16-builder), and also add new one (pg16.4-builder)?

alexeyklyukin (approved):
Looks good!

Makefile
graveland: Yeah. When using the builder in a FROM:, you'd generally want to use the `pg16-builder` tag to get the latest all the time. You'd only want to use the 16.4-builder tag when you're doing something that specifically requires a point release.

graveland: Yeah, it's a legacy variable name- changed it to VAR_PGMAJORMINOR

PR #490, opened 2024-09-12
Make sure to depend on the version info

graveland:


Makefile
```diff
@@ -262,6 +262,7 @@ build:
 
 .PHONY: publish-combined-builder-manifest
 publish-combined-builder-manifest: # publish a combined builder image manifest
+publish-combined-builder-manifest: $(VAR_VERSION_INFO)
```
minkimipt, line 265: This means that this target depends on the existence of `VAR_VERSION_INFO` variable. Are there cases when that variable is not set?  I see it's not defined as a dynamic variable, but maybe I don't get something here?
https://github.com/timescale/timescaledb-docker-ha/blob/brent/fix_builder_manifests/Makefile#L100

graveland: line 100 sets it to `version_info...` which is covered by the target defined at line 228.

It's complicated enough that it's not obvious exactly where/when the variables are set. I wouldn't be sad if we decided to use something other than a Makefile here. I wouldn't be surprised if this doesn't work, but this PR is a next step to test!

minkimipt (approved):

Makefile
minkimipt: Thanks. Now I feel I'm a tiny bit closer to understanding how it works.

graveland: Tested and it worked :)

PR #493, opened 2024-10-01
Set STOPSIGNAL to SIGINT

### LINKED ISSUE #492: SIGINT as the default stopsignal
This directly parallels [this issue](https://github.com/docker-library/postgres/issues/714) for the official docker postgres image, which was changed to [set SIGINT as the default stopsignal](https://github.com/docker-library/postgres/blob/a37e929682e8de45a3304a5bf9d63210c2e0a680/17/bookworm/Dockerfile#L194).

This would make the container use fast shutdown mode when stopped instead of trying to wait for all existing queries to finish and then being SIGKILL-ed 30 seconds later

graveland:
See: https://github.com/docker-library/postgres/issues/714

Fixes #492

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -547,6 +547,37 @@ ENV PGROOT=/home/postgres \
     LANG=C.UTF-8 \
     PAGER=""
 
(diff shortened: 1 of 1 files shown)


alexeyklyukin (approved):
Looks good!

PR #497, opened 2024-10-12
Refactor image introspection and tagging

graveland:
We now directly pass the builder and release IDs to the subsequent steps instead of depending on the names. We were accidentally using the previous builder when we were determining versions to come up with tag names.

This also adds a step to remove some of the contents of the github runner installations to make more room for building. /mnt has a lot of space but since we're doing all the work in docker, the root filesystem was sometimes filling up.

We also drop a docker install for arm images since the non-beta arm runners already include it now.

### DIFF
--- a/.github/workflows/publish_images.yaml
+++ b/.github/workflows/publish_images.yaml
@@ -46,6 +46,15 @@ jobs:
     runs-on: "${{ matrix.runs_on }}"
 
     steps:
+      # The github runners have a lot of space in /mnt, but apparently not enough in /. This step removes about 13G.
+      - name: remove unneeded runner software
+        run: |
+          df -h
+          du -chs     /usr/share/dotnet /usr/local/lib/android /opt/microsoft || true
+          sudo rm -fr /usr/share/dotnet /usr/local/lib/android /opt/microsoft || true
+          sudo docker image prune --all --force || true
+          df -h
+
       - name: Checkout code
         uses: actions/checkout@v4
 
@@ -64,6 +73,7 @@ jobs:
           endpoint: ha-builder
 
       - name: Build and publish (pg${{ matrix.pg_major }}${{ matrix.all }}${{ matrix.oss }} ${{ matrix.platform }})
+        id: build
         env:
           PLATFORM: ${{ matrix.platform }}
           PG_MAJOR: ${{ matrix.pg_major }}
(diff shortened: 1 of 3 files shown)


.github/workflows/publish_images.yaml
```diff
@@ -46,28 +46,18 @@ jobs:
     runs-on: "${{ matrix.runs_on }}"
 
     steps:
+      # The github runners have a lot of space in /mnt, but apparently not enough in /. This step removes about 13G.
+      - name: remove unneeded runner software
+        run: |
+          df -h
+          du -chs     /usr/share/dotnet /usr/local/lib/android /opt/microsoft || true
+          sudo rm -fr /usr/share/dotnet /usr/local/lib/android /opt/microsoft || true
```
alexeyklyukin, line 54: 💯

Makefile
```diff
@@ -94,10 +94,40 @@ INSTALL_METHOD?=docker-ha
...
-VAR_PGMAJORMINOR="$$(awk -F '=' '/postgresql.version=/ {print $$2}' $(VAR_VERSION_INFO) 2>/dev/null)"
-VAR_TSVERSION="$$(awk -F '=' '/timescaledb.version=/ {print $$2}' $(VAR_VERSION_INFO) 2>/dev/null)"
-VAR_TSMAJOR="$$(awk -F '[.=]' '/timescaledb.version=/ {print $$3 "." $$4}' $(VAR_VERSION_INFO))"
-VAR_VERSION_INFO=version_info-$(PG_MAJOR)$(DOCKER_TAG_POSTFIX).log
+VERSION_INFO = /tmp/outputs/version_info
+VAR_PGMAJORMINOR="$$(awk -F '=' '/postgresql.version=/ {print $$2}' $(VERSION_INFO) 2>/dev/null)"
+VAR_TSVERSION="$$(awk -F '=' '/timescaledb.version=/ {print $$2}' $(VERSION_INFO) 2>/dev/null)"
+VAR_TSMAJOR="$$(awk -F '[.=]' '/timescaledb.version=/ {print $$3 "." $$4}' $(VERSION_INFO))"
+
+# In these steps we do some introspection to find out some details of the versions
```
alexeyklyukin, line 102: Thank you for the comments!

JamesGuthrie (approved):

PR #514, opened 2025-01-20
feat: upgrade pgai extension to 0.7.0

smoya:


Dockerfile
```diff
@@ -149,6 +149,9 @@ RUN set -eux; \
 # We install pip3, as we need it for some of the extensions. This will install a lot of dependencies, all marked as auto to help with cleanup later
 RUN apt-get install -y python3 python3-pip
 
+# Upgrade pip to 24.3.1 to match the version in the non-ha timescaledb-docker image.
+RUN python3 -m pip install --upgrade pip==24.3.1
```
smoya, line 153: Matches with https://github.com/timescale/timescaledb-docker/pull/291/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R78

graveland: Is there any specific reason to want to upgrade pip? I'm not against it, just curious.

graveland (approved):
lgtm

Dockerfile
smoya: pgai Dockerfile command runs `pip install --break-system-packages uv==0.5.20` . [See here](https://pip.pypa.io/en/stable/news/#v23-0-1).  `--break-system-packages` is only available in pip >= 23.0.1. As in the non-ha docker image the package installed by alpine linux is 24.3.1, I'm fixing it the same here per compatibility.

graveland: Sounds good, thanks!

JamesGuthrie: I don't think that this is necessary.

JamesGuthrie (changes requested):

Dockerfile
JamesGuthrie: Oh I see why this is necessary now 🙈  obviously it's my fault. Chesterton's fence and all.

JamesGuthrie: For completeness: in https://github.com/timescale/pgai/commit/3f9736a0c3d91b5891395f47a41a1c27859ce244 we added support for using uv to install dependencies. As part of that change, I (effectively) replaced `PIP_BREAK_SYSTEM_PACKAGES=1 pip` with `pip --break-system-packages`. The advantage of the former is that it is a backwards-compatible method of telling pip to break system packages.

JamesGuthrie (approved):

Dockerfile
smoya: BTW, I changed the script to only upgrade if pip version is lower than `23.0.1`, which is the version that introduced the `--break-system-packages` flag. That way, the script is aware of the current version from now on and less intrusive.

PR #518, opened 2025-01-28
fix pg12 build

JamesGuthrie:
The latest pg12 packages contain a Build-Depends on an as-yet unreleased
package: postgresql-common-dev. This package should be released in mid
Febrary 2025. In the meantime we work around this by creating an empty
package.

See: https://www.postgresql.org/message-id/flat/D6BD9F9D-8DBE-4E0B-B00A-2130E7C6385D%40guthrie.ch


graveland (approved):
lgtm. Could you update the comment to mention it's temporary and when we should remove it too please?

PR #519, opened 2025-01-29
Use uv to install pgai dependencies

JamesGuthrie:
pgai's `build.py install` will automatically use uv if it's available on the system path. When uv installs dependencies into a target directory, it uses hardlinks to a centralized cache. This allows a large number of dependencies to be reused across pgai versions.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -149,11 +149,8 @@ RUN set -eux; \
 # We install pip3, as we need it for some of the extensions. This will install a lot of dependencies, all marked as auto to help with cleanup later
 RUN apt-get install -y python3 python3-pip
 
-# pgai requires pip 23.0.1 or greater due to the use of --break-system-packages flag
-RUN set -ex; \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -149,10 +149,10 @@ RUN set -eux; \
 # We install pip3, as we need it for some of the extensions. This will install a lot of dependencies, all marked as auto to help with cleanup later
 RUN apt-get install -y python3 python3-pip
 
-# pgai requires pip 23.0.1 or greater due to the use of --break-system-packages flag
+# using uv with pgai reduces size of dependencies
 RUN set -ex; \
     if [ "$(pip --version | awk '{print $2; exit}')" \< "23.0.1" ]; then \
```
graveland, line 154: This conditional should check uv instead of pip, possibly `if ! which uv >&/dev/null`?

JamesGuthrie: I didn't intend to leave the conditional in - dropped it entirely.

PR #535, opened 2025-03-26
feat: Add pg_semver extension to image

saolof:
pg_semver is from the same author as pg_tap and adds a semantic version type to postgres, and is distributed as part of the postgres apt repository.

It is technically possible to use a string column for this by enabling natural sorting at the unicode locale level with a collation, and by enabling regex validation to prevent git hashes from polluting the column and from introducing worst case times for unicode natural sorting. But pg_semver's representation is much more efficient and easier to introduce & enforce in practice if the extension is easily available.

This does add a small amount of bloat to the image (self-contained .so file and the plpgsql glue), so feel free to reject for any reason.


graveland (approved):
lgtm

PR #545, opened 2025-04-24
Ensure pgai files are owned by postgres

feikesteenbergen:
Certain files in the postgres lib and extension directories are owned by postgres:postgres to allow hot-patching of these files inside a container.

This wasn't true yet for pgai, by moving the pgai build commands into the block done by USER postgres, we should have files owned by postgres:postgres

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -201,19 +201,6 @@ RUN set -ex; \
         done; \
     fi
 
-# pgai is an extension for artificial intelligence workloads
-ARG PGAI_VERSION
-RUN set -ex; \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -317,6 +304,20 @@ USER postgres
 
 ENV MAKEFLAGS=-j4
 
+# pgai is an extension for artificial intelligence workloads
+ARG PGAI_VERSION
```
feikesteenbergen, line 308: Notice line 303, this is in a `USER postgres` context, instead of previously being in `USER root`

graveland (approved):

PR #550, opened 2025-05-21
Initial build for pg18

graveland:
This is obviously broken, as it doesn't even include timescaledb yet, but it should allow developers to start working on pg18.

It publishes to `timescaledev/timescaledb-ha` so it doesn't interfere with regular builds.


feikesteenbergen (approved):
I'd really like PG18 support sooner rather than later, so if it compiles, I'd merge it?

graveland:
> I'd really like PG18 support sooner rather than later, so if it compiles, I'd merge it?

That was the plan! Thanks :)

PR #556, opened 2025-06-02
fix: ensure PG version match is based upon major version only

thedodd:
This fixes a bug observed in downstream builds where version checks against, say PG17.4, fail due to 17.4 being greater than pgmax=17. The test is correct, but the semantic meaning is lost, and thus produces incorrect results.

With this changeset, we ensure that version checks for PG are based only upon MAJOR version for pgmin & pgmax as declared in our versions.yaml.

### DIFF
--- a/build_scripts/shared_versions.sh
+++ b/build_scripts/shared_versions.sh
@@ -103,6 +103,7 @@ install_rust_extensions() {
 
 version_is_supported() {
     local pkg="$1" pg="$2" ver="$3" pdata pgmin pgmax arch
+    pg="$(major_version_only "$pg")"
     local -a pgversions
 
     pdata="$(yq ".$pkg | pick([\"$ver\"]) | .[]" <<<"$VERSION_DATA")"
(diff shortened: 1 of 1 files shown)


build_scripts/shared_versions.sh
```diff
@@ -134,6 +137,14 @@ version_is_supported() {
     fi
 }
 
+# Ensure PG version matching is performed only based upon MAJOR version.
+#
+# This is a bit simplistic, but the system will otherwise break when
+# building a version of PG which specifies a minor or patch version.
+major_version_only() {
+    "$(awk -F'.' '{print $1}' <<<"$1")"
```
graveland, line 145: nit / not important, but you could drop a fork with native bash functionality:
```suggestion
    echo "${1%%.*}"
```

graveland (approved):

build_scripts/shared_versions.sh
thedodd: Thanks Brent! Done.

PR #587, opened 2025-09-10
add pg 18 to major 17 imae

sebastianwebber:
this pr adds pg 18 to the major 17 release.


graveland (approved):
As long as it successfully builds an image, I'd be okay with this.

PR #595, opened 2025-09-27
Add Tapir v0.0a to Docker image

tjgreen42:
## Summary
- Integrates Tapir BM25 index extension v0.0a into the TimescaleDB Docker image
- Currently supports PostgreSQL 17 only with both amd64 and arm64 architectures
- Excluded from OSS-only builds as Tapir is closed source for now


build_scripts/shared_install.sh
```diff
@@ -244,4 +244,47 @@ install_pgvectorscale() {
...
+
+            rm -rf /build/tapir
+            mkdir /build/tapir
+            cd /build/tapir
+
+            curl --silent \
+                 --fail \
+                 --location \
+                 --output artifact.zip \
+                 "https://github.com/timescale/tapir/releases/download/$version/tapir-$version-pg${pg}-${arch_deb}.zip"
```
graveland, line 284: If this is closed-source, can a !timescale person build this image, or will this cause a problem?

PR #653, opened 2026-04-15
fix: chown timescaledb.control to postgres so hot-forge can hot-patch it

minkimipt:
## Summary

- The extension directories are set up with sticky bit (`1775`, group=`postgres`), which prevents the postgres user from removing files owned by root even though the directory is group-writable
- hot-forge replaces files by first calling `remove_file` then creating a symlink; for `.so` files this works because they are new version-named files, but `timescaledb.control` is a fixed-name file installed as `root:root` by the loader apt package
- After the hot-forge bundle install, `default_version` in `pg_available_extensions` would stay at the image-baked version instead of reflecting the hot-patched version
- This adds a `chown postgres` step after installing the timescaledb packages, so hot-forge can remove the file and replace it with a symlink to its live directory

## Test plan

- [ ] Build a new image and verify `timescaledb.control` is postgres-owned: `stat /usr/share/postgresql/18/extension/timescaledb.control`
- [ ] Install a hot-forge bundle with a newer timescaledb version and verify `SELECT default_version FROM pg_available_extensions WHERE name = 'timescaledb'` returns the hot-patched version

Related PR https://github.com/timescale/hot-forge/pull/424

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -405,6 +405,17 @@ RUN OSS_ONLY="${OSS_ONLY}" \
         TIMESCALEDB_VERSIONS="${TIMESCALEDB_VERSIONS}" \
         /build/scripts/install_extensions timescaledb
 
+# The extension directories use a sticky bit (1775), which prevents the postgres user from
+# removing or renaming files owned by root — even when the directory itself is group-writable.
+# hot-forge replaces files by first removing them and then creating a symlink; for that to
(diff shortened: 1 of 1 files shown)


minkimipt:
Moving this fix to the private cloud image repo instead — hot-forge is cloud-internal and the change doesn't belong in the public OSS image.

PR #657, opened 2026-04-29
chore: pin github-actions to their SHAs, add dependabot config

nickstamat:


.github/workflows/build_branch.yaml
```diff
@@ -46,16 +46,16 @@ jobs:
           df -h
 
       - name: Checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
```
inchuev, line 49: I think there's checkout v6 already - why not switch to the latest?

inchuev (approved):
Other deps could also be updated?

.github/workflows/build_branch.yaml
nickstamat: The arm64 build runs on runners where I have not yet confirmed that checkout v6 is supported.
For this reason, I focused solely on reproducibility in this PR (i.e. the versions are preserved but are now pinned).

In any case, Dependabot is going to open PRs to update them. We'll then review & validate in those.

PR #668, opened 2026-06-08
chore: include toolkit latest release

adolsalamanca:
- Bumps Rust to 1.90.0 and cargo-pgrx to 0.18.0 to include the latest toolkit release.
- Fixed a broken version-detection regex by quoting it, stopping cargo-pgrx 0.18.0 from being misidentified as legacy, which broke system Postgres provisioning.
- Deferred adding pg_textsearch to shared_preload_libraries until after it is installed, preventing initdb from failing on pg17/18 due to a missing library.


minkimipt (approved):
I understand the intention of this PR, but I might be missing a lot of details of how things need to work. Approving it for the lack of ideas how else to check it.

surister (approved):
Lgtm

PR #672, opened 2026-06-12
Cleanup the runner on check steps too

graveland:
We were cleaning up on build steps, but not check steps, and it's close enough to cause intermittent job failures.

### DIFF
--- a/.github/workflows/publish_images.yaml
+++ b/.github/workflows/publish_images.yaml
@@ -176,6 +176,17 @@ jobs:
         docker_tag_postfix: ["", "-all", "-oss", "-all-oss" ]
 
     steps:
+      # Free up disk space on the runner. The check pulls both the amd64 and
+      # arm64 variants of the image, and the pg*-all images in particular are
+      # large enough that a stock runner's ~14GB free on / overflows otherwise.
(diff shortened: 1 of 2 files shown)


.github/workflows/publish_images.yaml
```diff
@@ -176,6 +176,17 @@ jobs:
         docker_tag_postfix: ["", "-all", "-oss", "-all-oss" ]
 
     steps:
+      # Free up disk space on the runner. The check pulls both the amd64 and
+      # arm64 variants of the image, and the pg*-all images in particular are
+      # large enough that a stock runner's ~14GB free on / overflows otherwise.
+      - name: remove unneeded runner software
+        run: |
+          df -h
+          sudo rm -fr /usr/share/dotnet /usr/local/lib/android /opt/microsoft \
```
murrayju, line 185: Don't think I've ever seen anyone use `-fr` over `-rf` 😂

murrayju (approved):

.github/workflows/publish_images.yaml
graveland: huh... never thought too much about it, but it's definitely the order ingrained in my head :)

PR #695, opened 2026-09-01
Shrink the builder image

nateships:
## Summary

The `pgNN-all-builder` image is 4.97 GB compressed. This PR cuts it to 3.39 GB (-32%) and keeps the layer structure. The files that downstream builds use do not change.

## Why the image is large

Image layers are copy-on-write. When a later `RUN` changes the mode or owner of a file from an earlier layer, Docker copies the whole file into the new layer. Three such steps hold 815 MiB of duplicate data, 18% of the image:

| Layer | Command | Size |
|---|---|---|
| 62 | `chmod -x /usr/lib/postgresql/*/lib/*.so` | 565 MiB |
| 43 | `chgrp -R postgres /usr/lib/debug` | 170 MiB |
| 42 | `chown -R postgres /usr/local/cargo` | 80 MiB |

A delete in a later layer has the same problem: the bytes stay in the lower layer.

## Changes

1. **Remove build trees after install, in the same step.** `/build/toolkit` (1.7 GB, mostly the cargo `target/` directory) and `/build/timescaledb` (0.2 GB) stay in the image today. The pgvectorscale and pg_textsearch installs already remove their trees. This PR does the same for toolkit and timescaledb, inside `install_extensions`.

2. **Remove `/usr/local/cargo/registry/src`** (0.95 GB) in the same step that builds the rust extensions. Cargo recreates these files from `registry/cache` when a build needs them.

3. **Fix the three copy-on-write steps at the source.**
   - `chown` of `/usr/local/cargo` moves into the `RUN` that installs rust.
   - `/usr/lib/debug` gets `chgrp` and `g+w` on the directories only. The dbgsym files stay in their own layer. Postgres can still add files to those directories.
   - `install_extensions` runs `chmod -x` on the `.so` files it installed, in the layer that created them. The final `chmod -x` in the Dockerfile then finds nothing to change. Coreutils skips files that already have the target mode, so a no-op chmod costs no layer space.

## Result

Measured on the images that the `publish_check` run for 3773a80 pushed ([run 33896369024](https://github.com/timescale/timescaledb-docker-ha/actions/runs/33896369024)). The run also passes the builder smoketest.

| Image (amd64) | Layers | Compressed |
|---|---|---|
| `pg18-all-builder-amd64` (master) | 68 | 4.97 GB |
| `pg18-cicheck-all-builder-amd64` (this PR) | 67 | 3.39 GB |

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,12 @@
 ## tools we use will be the same across the board, as most of our tools our
 ## installed using external repositories.
 ARG DOCKER_FROM=ubuntu:22.04
-FROM ${DOCKER_FROM} AS builder
+ARG RUST_VERSION=1.96.0
+ARG BUILD_PACKAGES="binutils cmake devscripts equivs gcc git gpg gpg-agent libc-dev libc6-dev libkrb5-dev libperl-dev libssl-dev lsb-release make patchutils python2-dev python3-dev wget libsodium-dev"
+
+## The build-stage does all the work. The published `builder` target below
+## copies its filesystem into a single layer.
+FROM ${DOCKER_FROM} AS build-stage
 
 SHELL ["/bin/bash", "-exu", "-o", "pipefail", "-c"]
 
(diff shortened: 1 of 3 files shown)


minkimipt (approved):

Dockerfile
```diff
@@ -520,7 +530,42 @@ USER postgres
 COPY --chown=postgres:postgres cicd /cicd/
 RUN /cicd/install_checks -v
 
-FROM builder AS trimmed
+## The published builder image is one layer. A chown or chmod on a file from
+## an earlier layer copies that file into a new layer; one layer avoids the
+## duplicates. The image config below must match what build-stage sets.
+FROM scratch AS builder
+COPY --from=build-stage / /
```
surister: if we just copy everything from earlier stages, what's the purpose of using multi-stage images, In my head, you copy what you need from image A to B, and that's why B is small, if we blank copy, isn't it the same as just having A?

nickstamat (approved):

PR #54, opened 2020-06-26
Feike/rename tags

feikesteenbergen:
# Upgrade tsdb_admin

# Publish Docker Image under different tags

After some discussion, we decided to only publish mutable images with
the `latest` postfix, as this clearly communicates that the tag is
immutable.

Together with that, we also publish 1 immutable tag for every git tag in
our repository.

PR #698, opened 2026-09-05
Build each extension version in its own layer, on runs-on runners with layer caching

nateships:
## Summary

Stacked on #695. Builds run on runs-on runners with a layer cache. Each extension version has its own layer, so a version bump rebuilds only that layer.

## Changes

- One layer per extension version, generated from `build_scripts/versions.yaml`. Recent timescaledb minors get one layer each, older versions share one. `versions.yaml` is copied in after these layers, so a change to it does not rebuild them.
- Build jobs run on the on-demand 16 cpu runs-on fleets.
- Layer cache in the runs-on S3 bucket, one manifest per image variant, platform and branch. A build reads its own and master's.
- Packages are not pinned. The weekly scheduled publish builds cold and refreshes them. A publish from master ships what the PR build tested.
- Stock apt sources. The AWS mirror override is gone.
- runs-on workarounds: the smoketest runs the pushed image by digest (the ECR mirror can serve a stale tag) and with `--security-opt apparmor=unconfined` (PostgreSQL 18 hangs without it, see timescale/timescaledb-docker-cloud#209).
- `install_checks` also checks the image layout: library modes, removed build trees, no rust toolchain in the release.

## Result

pg18-all, one job each:

| Platform | Before | Cold | Cached |
|---|---|---|---|
| amd64 | 180 min | 69 min | 7 min |
| arm64 | 160 min | 72 min | 2 min |

A cached run is a build with no version change. The published files do not change.

## How to add or change a version

1. Edit `build_scripts/versions.yaml`.
2. Run any image build, or `make dockerfile`, and commit the Dockerfile with it. CI fails when the committed Dockerfile is out of date.
3. A new PostgreSQL minor goes in `build_scripts/postgres_versions.yaml`. It rebuilds everything after the PostgreSQL install.
4. Branch builds (`TIMESCALEDB_VERSIONS=main`, `feature/x`) need no Dockerfile change.
5. `mise install` gives you `yq`, `jq` and `shellcheck`.


minkimipt:
One thing worth folding into this PR while you're in here: `docker builder prune --force`
still runs between `publish-builder` and `publish-release`, three lines below the new
`DOCKER_CACHE_SCOPE`/`DOCKER_CACHE_BRANCH` env — in both `publish_check.yaml` and
`publish_images.yaml`. As it stands the job sets up a cache and then throws it away
mid-build.

The disk-pressure rationale it was added for doesn't hold on the RunsOn fleet. Full `df`
from a complete pg18-all run yesterday, both samples the step itself prints:

```console
Filesystem       Size  Used Avail Use% Mounted on
/dev/root        193G   17G  177G   9% /
/dev/root        193G   11G  182G   6% /
```

Never above 9% of 193G. That was a real constraint on `ubuntu-22.04` (~14G free), not here.

Same applies to the `remove unneeded runner software` step it sits next to (4 sites across
the three workflows) — it also `rm -fr /opt/hostedtoolcache` while RunsOn sets
`RUNNER_TOOL_CACHE=/opt/hostedtoolcache`, so it deletes the directory the runner advertises
as its tool cache. Harmless today since nothing uses `setup-*`, but it'll bite whoever adds one.

Happy to send these as a follow-up PR instead if you'd rather keep this diff focused — flagging
here mainly so the prune doesn't quietly negate the caching once this merges. For context on
why it matters: `#55` (toolkit, all versions) and `#54` (timescaledb) are 46.5 min and 21.1 min
of a ~100 min run, so the cache is most of the win available.

adolsalamanca (approved):

PR #699, opened 2026-09-05
Build on runs-on 16 cpu runners with a shared layer cache

nateships:
## Summary

Stacked on #698. Moves the image builds to runs-on fleet runners and shares the BuildKit layer cache between builds through the runs-on S3 cache bucket.

## Changes

1. **Runners.** The build jobs in `publish_images`, `publish_check`, and `build_branch` run on runs-on fleet runners (`runs-on/fleet=<name>/env=ue1`):

   | Workflow | amd64 | arm64 |
   |---|---|---|
   | `build_branch`, `publish_check` | `linux-16cpu-x64` | `linux-16cpu` |
   | `publish_images` | `linux-16cpu-x64-ondemand` | `linux-16cpu-ondemand` |

   Publish uses the on-demand fleets so a spot interruption cannot cost a 3 hour build. The `runs-on/action` step is added before checkout. The `linux-16cpu-x64` and the two on-demand fleets are new in the runs-on config (savannah-infra, branch `runson-ubuntu24-shapes`) and must be applied before this merges.

2. **Layer cache.** The Makefile adds `--cache-from` and `--cache-to type=s3,...` when `DOCKER_CACHE_SCOPE` is set. Bucket, prefix and region come from the `RUNS_ON_S3_BUCKET_CACHE`, `RUNS_ON_S3_CACHE_REPO_PREFIX` and `RUNS_ON_AWS_REGION` variables that `runs-on/action` exports; the runner's IAM role grants access, so no token or extra action is needed. Each image variant and platform gets its own cache manifest (`name=pg18-all-amd64`), the blobs are shared.

3. **Cache policy.**
   - `build_branch` and `publish_check` read and write the cache.
   - `publish_images` only writes it (`DOCKER_CACHE_FROM=false`). BuildKit caches a `RUN apt-get update; apt-get upgrade` layer by its text, not its result. A publish that read the cache would ship stale packages until someone edited the Dockerfile above that line. The weekly publish exists to pick up upstream updates, so it stays cold and seeds fresh layers for the following PR builds.

## Expected effect

With #698, a version bump PR rebuilds only the new version's layer. Everything else comes from the cache written by the last publish. Cold publish builds stay long but run on 16 cpu instead of 4.

## To verify

- The `publish_check` run on this PR proves the fleet labels and the cache export. A second push to the branch should then show the extension layers as `CACHED`.
- Fork PRs already need approval before any workflow runs (`fork-pr-contributor-approval` is `all_external_contributors` on the repo), so no fork reaches these runners without a maintainer approving the run.

## Follow-ups outside this repo

- Apply the runs-on config change that adds the `16cpu-linux-x64` shape, the 16 cpu on-demand shapes, and the `linux-16cpu-x64`, `linux-16cpu-ondemand`, `linux-16cpu-x64-ondemand` fleets (200 GB volumes).

### DIFF
--- a/.github/workflows/build_branch.yaml
+++ b/.github/workflows/build_branch.yaml
@@ -30,11 +30,13 @@ jobs:
         platform: [ amd64, arm64 ]
         include:
           - platform: amd64
-            runs_on: ubuntu-22.04
+            runs_on: runs-on/fleet=linux-16cpu-x64/env=ue1
           - platform: arm64
-            runs_on: cloud-image-runner-arm64
+            runs_on: runs-on/fleet=linux-16cpu/env=ue1
 
     steps:
+      - uses: runs-on/action@46910bf61b41721b0579f237e186afb35477007a # v2.3.0
+
       # Free up disk space on the runner. The pg18-all builds in particular need
       # a lot of space due to 4 PG versions worth of extensions and debug symbols.
       - name: remove unneeded runner software
@@ -64,6 +66,7 @@ jobs:
       - name: Build
         env:
           PLATFORM: ${{ matrix.platform }}
+          DOCKER_CACHE_SCOPE: pg${{ env.PG_MAJOR }}-all-${{ matrix.platform }}
         run: make build-sha
 
       - name: Check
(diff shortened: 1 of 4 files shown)


nateships:
Folded into #698 so the layer split and the cache can be tested together.

